PR #11, opened 2018-02-01
Multi-file download and initial tests

nataliemona:


functions/zoom-downloader.py
```diff
@@ -223,3 +156,76 @@ def send_sqs_message(record):
...
+    link = source_object['src']
+
+    return link
+
+
+def get_connection_using_play_url(file):
+    file_url = retrieve_url_from_play_page(file['play_url'])
+    zoom_name = file_url.split("?")[0].split("/")[-1]
+
+    r = requests.get(file_url, stream=True)
```
lbjay, line 198: Was just reading about the details of `stream=True` and I think it might be good to make sure these response objects get `close()`ed at some point http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow

They recommend using a `with...` context but that might require too much refactoring so just a `r.close()` call somewhere is fine.

lbjay (approved):

lbjay:
Merging this. The one thing i raised can be dealt with later.

PR #16, opened 2018-02-12
Switched DynamoDB download tables to SQS queue

nataliemona:


nataliemona:
Downloader now only has 1 test since most became irrelevant with the switch to SQS.

functions/zoom-downloader.py
```diff
@@ -38,24 +35,22 @@ def handler(event, context):
...
         return
 
-    record = json.loads(event['Records'][0]['dynamodb']['NewImage']['recording_data']['S'])
-    logger.debug(record)
+    download_data = json.loads(message.body)
+    logger.info(download_data)
+    recording_data = download_data['recording_data']
 
-    chronological_files = sorted(record['recording_files'], key=lambda k: k['recording_start'])
+    chronological_files = sorted(recording_data['recording_files'], key=lambda k: k['recording_start'])
```
lbjay: Another useful approach to this:
```
from operator import itemgetter
...
sorted(recording_data['recording_files'], key=itemgetter('recording_start'))
```
Not really better in this case, but in a situation where performance was a big deal.

functions/zoom-downloader.py
```diff
@@ -68,11 +63,12 @@ def handler(event, context):
         logger.debug("file {} is track sequence {}".format(file, track_sequence))
         prev_file = file
 
-        stream_file_to_s3(file, record['uuid'], track_sequence)
+        stream_file_to_s3(file, recording_data['uuid'], track_sequence)
 
-    record['downloader_correlation_id'] = context.aws_request_id
-    send_sqs_message(record)
+    download_data['downloader_correlation_id'] = context.aws_request_id
```
lbjay: i think this should be `recording_data`

PR #19, opened 2018-03-12
Relocate fetching of recording data to the downloader function

lbjay:
* webhook now limited to fetching meeting host data
* simplified some of the error condition logic in the recording data
  request. request will now be retried only on connection/timeout errors
  and recording status issues.


functions/zoom-downloader.py
```diff
@@ -38,23 +60,43 @@ def handler(event, context):
...
-        logger.debug("Messages received from download queue: {}".format(messages))
-        message = messages[0]
-        logger.debug({'queue_message': message})
+        messages = download_queue.receive_messages(MaxNumberOfMessages=1)
+        download_message = messages[0]
+        logger.debug({'queue message': download_message})
 
     except IndexError:
-        logger.warning("No uploads ready for processing")
+        logger.info("No recordings ready for processing")
```
nataliemona: It would probably be more accurate to say "No download queue messages received" since no recordings ready is not the only reason that no messages would be received.

template.yml
```diff
@@ -72,8 +73,10 @@ Resources:
     Properties:
       QueueName: !Sub "${AWS::StackName}-downloads.fifo"
       FifoQueue: true
+      DelaySeconds: 300
       MessageRetentionPeriod: 1209600 # 14 days
       ContentBasedDeduplication: true
+      VisibilityTimeout: 700
```
nataliemona: I don't think setting the VisibilityTimeout in the template is a good idea because it will cause us to disrupt everything if we poll the SQS messages in the console.

nataliemona (approved):

PR #21, opened 2018-03-19
Task to move dead SQS messages back to source

nataliemona:


tasks.py
```diff
@@ -428,3 +450,76 @@ def _set_debug(ctx, debug_val):
                ).format(profile_arg(), new_vars, func_name)
         ctx.run(cmd)
 
+
+def _move_messages(queue_type, limit):
+    if queue_type == "uploads":
```
lbjay: can you simplify all this to:

    source_name = "{}-{}.fifo".format(env('STACK_NAME'), queue_type)
    dl_name = "{}-{}-deadletter.fifo".format(env('STACK_NAME'), queue_type)

You don't need to protect against invalid queue types because the `@task` defs control how this gets called.

lbjay (changes requested):
Looks good. Next I'd also like to see tasks that simply list the contents of the DLQs

PR #23, opened 2018-03-28
Zoom url unsafe characters fix, DLQ view task, opencast, reduce opencast uploads (for test), bug fixes & log cleanup

nataliemona:


template.yml
```diff
@@ -512,3 +515,78 @@ Resources:
       Principal: events.amazonaws.com
       SourceArn: !GetAtt [ZoomNewLogGroupEventRule, Arn]
 
+  RecordingCompletedMetricFilter:
```
lbjay, line 517: Nice. These are great.

functions/zoom-uploader.py
```diff
@@ -96,10 +99,10 @@ def created(self):
         return self.data['start_time']
 
     @property
-    def created_local(self, tz='US/Eastern'):
+    def created_local(self, tz=LOCAL_TIME_ZONE):
```
lbjay: I'm wondering if it would be better to have both the `created` and `created_local` properites return datetime objects vs strings. then do the string formatting where they get used.

functions/zoom-uploader.py
```diff
@@ -108,8 +111,11 @@ def title(self):
 
     @property
     def description(self):
+        formatted_time = datetime.strftime(
```
lbjay: previous suggestion about datetime objs would spare this back and forth parse/unparse.

functions/zoom-uploader.py
```diff
@@ -124,27 +130,58 @@ def s3_prefix(self):
     def zoom_series_id(self):
         return self.data['meeting_number']
 
+    def series_id_from_schedule(self):
+
+        dynamodb = boto3.resource('dynamodb')
+        table = dynamodb.Table(CLASS_SCHEDULE_TABLE)
```
lbjay, line 132: do you have something that populates this table yet?

functions/zoom-uploader.py
```diff
@@ -124,27 +130,58 @@ def s3_prefix(self):
...
+        weekdays = ['M', 'T', 'W', 'R', 'F']
+        if zoom_time.weekday() > 4:
+            logger.debug("Meeting occurred on a weekend.")
+            return None
+        elif weekdays[zoom_time.weekday()] not in schedule['Days']:
+            logger.debug("No opencast recording scheduled for this day of the week.")
+            return None
+
+        scheduled_time = datetime.strptime(schedule['Time'], '%H:%M')
+        scheduled_time = zoom_time.replace(hour=scheduled_time.hour, minute=scheduled_time.minute)
```
lbjay: do you mean to set `scheduled_time` twice here? I'm not following.

functions/zoom-uploader.py
```diff
@@ -124,27 +130,58 @@ def s3_prefix(self):
...
+        timedelta = abs(zoom_time - scheduled_time).total_seconds()
+
+        minutes = 20
+        if timedelta < (minutes * 60):
```
lbjay: does the schedule indicate how long the class is? what if something happens and the recording starts 30m late? would it be possible to do something like:

    if class_start_time - 20m < zoom_time < class_end_time:
        ...

functions/zoom-uploader.py
```diff
@@ -124,27 +130,58 @@ def s3_prefix(self):
...
+            return schedule['opencast_series_id']
+        else:
+            logger.debug("Meeting started more than 15 minutes before or after opencast scheduled start time.")
```
lbjay: "20"

functions/zoom-uploader.py
```diff
@@ -251,11 +288,21 @@ def create_mediapackage(self):
...
+        def sorted_videos(vids):
+            return sorted(
+                    vids,
+                    key=lambda vid: vid.metadata['track_sequence']
+                   )
+
+        if self.speaker_videos is not None:
+            videos = sorted_videos(self.speaker_videos)
+        elif self.gallery_videos is not None:
+            videos = sorted_videos(self.gallery_videos)
```
lbjay, line 299: are speaker vs gallery videos mutually exclusive? I thought a recording could have both.

functions/zoom-webhook.py
```diff
@@ -55,10 +56,13 @@ def handler(event, context):
             "Handling not implement for status '{}'".format(payload['status'])
         )
 
+    now = datetime.strftime(datetime.today(), '%Y-%m-%dT%H:%M:%SZ')
```
lbjay: can this be in UTC?

functions/zoom-uploader.py
nataliemona: Good idea.

functions/zoom-uploader.py
nataliemona: I do but it is a script in a jupyter notebook. If we decide on a standard json format we can add it to the project?

functions/zoom-uploader.py
nataliemona: I was just breaking up one computation into two lines but it could be written better.

functions/zoom-uploader.py
nataliemona: We do not have duration in the current spreadsheet.

functions/zoom-uploader.py
nataliemona: Since we are currently doing singlestream only (opencast presenter view) we are picking one or the other. But if we know how to do multistream in opencast (do we need to change something with our opencast cluster?) we could include both.

functions/zoom-webhook.py
nataliemona: Sure. I didn't think about it. I was mainly including this to know how old deadletter messages are so I was thinking more in terms of days or weeks rather then hours.

functions/zoom-uploader.py
lbjay: :+1:

functions/zoom-uploader.py
lbjay: Got it. I'll add an issue for me to work out the details of the dual stream workflow.

lbjay (approved):

PR #26, opened 2018-04-27
Import course schedule from csv to dynamodb

nataliemona:


tasks.py
```diff
@@ -596,3 +611,78 @@ def _view_messages(queue_type, limit):
...
+    elif semester not in ["01", "02", "03"]:
+        print("Semester must be '01' for fall. '02' for winter, or '03' for summer.")
+        return
+
+    csv_file = open(csv_name, "r")
+    json_file = open(json_name, "w")
+
+    fieldnames = tuple(csv_file.readline().split(',')[:-1])
+
+    reader = csv.DictReader(csv_file, fieldnames)
```
lbjay: why are you passing fieldnames when DictReader will detect them for you from 1st line of the file?

tasks.py
```diff
@@ -596,3 +611,78 @@ def _view_messages(queue_type, limit):
...
+
+    data = {}
+
+    for row in reader:
+        del row[None]
```
lbjay: What is this doing? Is the `restkey` arg to DictReader something that would be useful?

tasks.py
```diff
@@ -596,3 +611,78 @@ def _view_messages(queue_type, limit):
...
+
+        course['opencast_series_id'] = opencast_series_id
+        course['opencast_subject'] = opencast_subject
+        course['zoom_series_id'] = zoom_series_id
+        course['Days'] = [x.strip() for x in course['Day'].split("/")]
+        del course['Day']
+
+        data[zoom_series_id] = course
+
+    json.dump(data, json_file)
```
lbjay: you never know when a human might want to read this. i always use `indent=2` unless i'm in some kind of legit bandwidth or storage constraint.

tasks.py
nataliemona: Updated so that DictReader automatically recognizes fieldnames but this issue remains. The rightmost column in the spreadsheet has some random notes in it and DictReader places this under the fieldname of empty string. Dynamo won't accept this key so I'm actively deleting it. Tried restkey/restval but this doesn't work as technically the extra notes are in the field with the empty name (maybe those are more for malformed csv?).

tasks.py
nataliemona: Fixed.

tasks.py
nataliemona: Fixed.

PR #29, opened 2018-05-09
CodeBuild, Webhook V2, fix metric namespace collision

nataliemona:


lbjay (changes requested):
This is mostly good. The only change I'd like before merging and moving on is to incorporate the code build role permissions into the main ZoomIngesterRole. I think all that entails is adding an assumeRole statement for codebuild and then updating the codebuild template resource to use ZoomIngesterRole as it's service role.

PR #32, opened 2018-05-30
exec functions, DEFAULT_SERIES_ID bug fix

nataliemona:


tasks.py
```diff
@@ -79,7 +84,22 @@ def codebuild(ctx, revision):
...
+    print("Waiting for codebuild to finish...")
+    while build_status != "FAILED" and build_status != "COMPLETED":
+        time.sleep(5)
+        res = ctx.run(cmd, hide='out')
+        new_status = json.loads(res.stdout)["builds"][0]["currentPhase"]
+        if new_status != build_status:
+            print(build_status)
+            build_status = new_status
+
+    print("Build finished with status {}".format(build_status))
```
lbjay, line 100: Works great. Nice idea.

tasks.py
```diff
@@ -20,6 +22,9 @@
 STACK_NAME = env('STACK_NAME')
 PROD_IDENTIFIER = "prod"
 NONINTERACTIVE = env('NONINTERACTIVE')
+ZOOM_API_KEY = env('ZOOM_API_KEY')
```
lbjay: Along the way I've tried to stick to using `getenv(...)` within the task code vs defining these at the top. The advantage being that `getenv` complains properly if a necessary value isn't set (or a default given). I forgot to set my zoom base api and got a non-helpful stacktrace.

lbjay (changes requested):

lbjay (approved):

PR #38, opened 2018-06-19
Fix error log -> SNS mechanism

lbjay:
* LogGroups are created by cfn template so log subscriptions should be too
* log notifications function handler no longer needs to manage creation
  of subscription filters + lambda permissions
* updated formatting of notification emails to render readable
  tracebacks
* added a test for the function handler


functions/zoom-log-notifications.py
```diff
@@ -1,46 +1,26 @@
...
-def events2sns(event, context):
-
         logger.info("processing subscribed events")
 
         # log event data comes base64 encoded & gzipped
-        raw_data = zlib.decompress(
-            base64.b64decode(event['awslogs']['data']),
-            wbits=16
-        )
+        raw_data = base64.b64decode(gzip.decompress(event['awslogs']['data']))
```
lbjay, line 23: I couldn't get the zlib method to work in a reverse test context, so switched to gzip which is more straightforward anyway.

tasks.py
```diff
@@ -574,7 +575,8 @@ def __create_or_update(ctx, op):
                 getenv("LOCAL_TIME_ZONE"),
                 sg_id,
                 subnet_id,
-                getenv("LAMBDA_RELEASE_ALIAS")
+                getenv("LAMBDA_RELEASE_ALIAS"),
+                getenv("LOG_NOTIFICATIONS_FILTER_LOG_LEVEL", required=False) or ""
```
lbjay, line 579: I **think** this little `getenv(...) or ""` trick will avoid the situation where an env var ends up with the string `"None"`.

PR #43, opened 2018-07-17
concat bug fix

nataliemona:
opencast concatenates files in alphabetical order
prepend track sequence number to file name in s3 bucket to ensure correct ordering
new file naming format: {md5 of uuid}/{track number}-{file id}.{file type}

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -236,8 +236,7 @@ def next_track_sequence(prev_file, file):
 
 def stream_file_to_s3(file, uuid, track_sequence):
 
-    metadata = {'uuid': uuid,
-                'track_sequence': str(track_sequence)}
+    metadata = {'uuid': uuid}
 
     if file['file_type'].lower() in ["mp4", "m4a", "chat"]:
         stream, zoom_name = get_connection_using_play_url(file['play_url'])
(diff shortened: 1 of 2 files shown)


functions/zoom-downloader.py
```diff
@@ -250,7 +249,9 @@ def stream_file_to_s3(file, uuid, track_sequence):
         stream, zoom_name = get_connection_using_download_url(file['download_url'])
 
     metadata['file_type'] = zoom_name.split('.')[-1]
-    filename = create_filename(file['id'], file['meeting_id'], zoom_name)
+    filename = create_filename("{}-{}".format(track_sequence, file['id']),
```
lbjay: Do you know how the concat operation sorts the filenames? I suspect it's lexical, which means this won't do what we want if the list of files is > 9 (probably unlikely but still).

```
In [1]: ids = [str(x) for x in range(1,11)]

In [2]: sorted(ids)
Out[2]: ['1', '10', '2', '3', '4', '5', '6', '7', '8', '9']
```

Could you change the integers to letters? Then we'd be safe for <= 26 files.

```
from string import ascii_uppercase
...
filename = create_filename("{}-{}".format(ascii_uppercase[track_sequence - 1], file['id']), ...
```

functions/zoom-uploader.py
```diff
@@ -341,16 +341,10 @@ def add_tracks(self):
...
 
-        def sorted_videos(vids):
-            return sorted(
-                    vids,
-                    key=lambda vid: vid.metadata['track_sequence']
-                   )
-
         if self.speaker_videos is not None:
-            videos = sorted_videos(self.speaker_videos)
+            videos = self.speaker_videos
```
lbjay, line 345: Is this because we assume the order in which the tracks are added doesn't matter now?

lbjay (changes requested):

functions/zoom-downloader.py
nataliemona: Instead of letters, changed to 3 digit ints. So 009, 010, 011, etc.

functions/zoom-uploader.py
nataliemona: Yes. Getting rid of unnecessary code and metadata.

lbjay (approved):

PR #45, opened 2018-08-09
match metadata to opencast standard

nataliemona:


functions/zoom-uploader.py
```diff
@@ -150,7 +150,7 @@ def series_id_from_schedule(self):
             schedule = r['Item']
             logger.info(schedule)
 
-        zoom_time = self.created.astimezone(timezone(LOCAL_TIME_ZONE))
+        zoom_time = self.created_local
```
lbjay: I feel like these time values are churning a lot?

functions/zoom-uploader.py
```diff
@@ -268,6 +268,7 @@ def upload(self):
             logger.warning("Episode with mediapackage id {} already ingested"
                            .format(self.mediapackage_id))
             return None
+        self.get_series_catalog()
```
lbjay, line 267: can it get to this point without a valid series id?

functions/zoom-uploader.py
```diff
@@ -300,6 +301,18 @@ def create_mediapackage_id(self):
...
 
+    def get_series_catalog(self):
+
+        logger.info("Getting series catalog for series: {}"
+                    .format(self.opencast_series_id))
+
+        endpoint = "/series/{}.json".format(self.opencast_series_id)
+        resp = oc_api_request('GET', endpoint)
+
+        logger.debug(resp.text)
```
lbjay: this should probably be something like, `logger.debug({'series_catalog': resp.text})`

functions/zoom-uploader.py
nataliemona: Yes, I agree...

functions/zoom-uploader.py
nataliemona: It's possible and then it would throw an error which is what we would want it to do.

lbjay (approved):

PR #47, opened 2018-09-28
Opencast 5.x compatibility

nataliemona:


tasks.py
```diff
@@ -282,7 +287,7 @@ def exec_uploader(ctx, series_id=None, ignore_schedule=False, qualifier=None):
     """
     Manually trigger uploader.
     """
-    payload = { 'num_uploads': 1, 'ignore_schedule': ignore_schedule }
+    payload = { 'ignore_schedule': ignore_schedule }
```
lbjay, line 290: It just occurred to me that an `exec.uploader` command will now trigger `UPLOAD_MESSAGES_PER_INVOCATION` number of queue messages, which is different from what we're used to. Not sure if that's a problem, but if so, the manually invocation task might need to have a way to override.

tasks.py
```diff
@@ -633,63 +638,112 @@ def __create_or_update(ctx, op):
...
+            cmd = ("aws {} cloudformation describe-change-set --change-set-name {} "
+                   "--output text --query 'StatusReason'").format(
+                profile_arg(),
+                change_set_id
+            )
+            ctx.run(cmd)
+        else:
+            print("\nCloudFormation stack changeset created.\n")
+            cmd = ("aws {} cloudformation describe-change-set --change-set-name {} "
+                   "--query 'Changes[*].ResourceChange.{{ID:LogicalResourceId,Change:Details[*].CausingEntity}}' "
```
lbjay, line 715: nice :+1:

lbjay (approved):

PR #48, opened 2018-10-05
moar 5x changes

lbjay:
Most important thing here is the change to the uploader message handling. I'm hoping 10 isn't too many, I figure the execution time / work it does to determine if something is ingestable is pretty minimal, like < 1s each to check the dynamo schedule. So if it gets to the 10th message and finds something that needs ingesting it's burned through 10s of it's allowed execution time? Seems fine to me.

I'm the king of mucking with stuff that isn't **THE** immediate issue, so of course I added some extra features I've been wishing for. I've very excited about the log search thing. I thought the cfn change-sets were going to be more informative, but at least it shows you what pieces of the stack will be affected by an update.

### DIFF
--- a/example.env
+++ b/example.env
@@ -31,6 +31,11 @@ PARALLEL_ENDPOINT=
 # id of the opencast opsworks cluster's vpc
 VPC_ID=
 
+# URL and API auth for the target opencast cluster
+OPENCAST_BASE_URL=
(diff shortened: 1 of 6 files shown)


lbjay:
Oops, I owe this PR some README changes. Next week.

PR #50, opened 2018-12-05
Select 1 MP4 per time segment, handle multiple subnets.

nataliemona:
Pick 1 MP4 per time segment based on the priority list: 'shared_screen_with_speaker_view', 'shared_screen', 'active_speaker.'
If no "PrivateSubnet" look for "PrivateSubnet1".

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -84,7 +84,9 @@ def handler(event, context):
         logger.info("Moved message to DLS and deleted message from source queue.")
         raise
 
-    chronological_files = sorted(recording_data['recording_files'], key=itemgetter('recording_start'))
+    chronological_files = sort_files(recording_data['recording_files'])
+    if not chronological_files:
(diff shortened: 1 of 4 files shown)


lbjay:
I pushed an additional commit here that reworks a bit the vpc subnet / security group stuff. I realized lately that i much prefer using the awscli `--query` param vs parsing json output.

functions/zoom-downloader.py
```diff
@@ -374,3 +376,45 @@ def is_valid_mp4(filename):
     else:
         logger.debug("Successfully verified mp4 {}".format(url))
         return True
+
+
+def sort_files(files):
```
lbjay: I hate to do this after sitting on the PR for so long, but I'm not loving how the filtering and sorting is lumped together. Is it too complicating to break that up into two steps?

lbjay (approved):

PR #53, opened 2019-02-08
exec.pipeline task for more efficient testing

nataliemona:


tasks.py
```diff
@@ -53,6 +53,18 @@ def get_queue_url(queue_name):
 
     return queue_url
 
+
+def empty_queue(ctx, queue_name):
```
lbjay: this should be `queue_is_empty`

lbjay (changes requested):

PR #55, opened 2019-02-28
Move schedule matching to downloader.

nataliemona:


functions/zoom-downloader.py
```diff
@@ -60,96 +68,213 @@ def handler(event, context):
...
+
+        return self.upload_message
+
+
+def meeting_metadata(uuid):
+    """
+    Get metadata about meeting. Metadata includes zoom meeting id,
+    host_name, host_email, topic, and start_time.
+    """
+    for meeting_type in ['past', 'pastOne', 'live']:
```
lbjay, line 265: It would be good if, somewhere in this chain of functions (get_api_data, api_request) each request attempt was logged at at least debug level.

lbjay: or maybe somewhere in the requests/urllib modules it does? if so, nevermind.

lbjay (approved):
:+1:

functions/zoom-downloader.py
nataliemona: Ran it to be sure, yes urllib logs the requests.

PR #57, opened 2019-03-26
fix log notifications

nataliemona:


tasks.py
```diff
@@ -437,18 +437,18 @@ def delete(ctx):
         ctx.run(delete_files, echo=True)
         ctx.run(delete_stack, echo=True)
         __wait_for(ctx, 'stack-delete-complete')
-    else:
-        print("not deleting stack")
```
nataliemona, line 440: Don't delete files if stack deletion cancelled.

functions/zoom-log-notifications.py
```diff
@@ -23,7 +23,7 @@ def handler(event, context):
 
         # log event data comes gzipped & base64 decoded
         raw_data = event['awslogs']['data']
-        decoded = codecs.decode(raw_data, 'base64', 'strict')
+        decoded = codecs.decode(bytes(raw_data, 'utf8'), 'base64', 'strict')
```
lbjay, line 26: ahhh! :man_facepalming:

lbjay (approved):

PR #59, opened 2019-09-05
initial ssm-dotenv setup (DON'T MERGE)

nataliemona:


tasks.py
```diff
@@ -646,24 +672,33 @@ def logs_uploader(ctx, watch=False):
...
-    return val
-
-
 def profile_arg():
     if AWS_PROFILE is not None:
         return "--profile {}".format(AWS_PROFILE)
     return ""
 
 
+def getenv(var, required=True):
```
lbjay, line 684: looks like this is defined twice

tasks.py
```diff
@@ -36,10 +36,36 @@
     'zoom-log-notifications'
 ]
 
+if PROD_IDENTIFIER in STACK_NAME and AWS_PROD_PROFILE is not None:
```
lbjay, line 40: can you walk me through how this would work in practice?

PR #63, opened 2019-11-18
Ignore short recordings

nataliemona:
- Filter out recordings from meetings that are < 2 minutes long.
- Updated all dependencies (and tested the updated version).
- Updated the Codebuild image (used to only support python 3.4 but new images have been released now https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html). Now using aws/codebuild/amazonlinux2-x86_64-standard:1.0
- Updated all lambda functions to use Python 3.7
- Did some code cleanup, installed a linter that got mad about line length... so ignore those changes... I added comments to the important changes.


function_requirements/zoom-downloader.txt
```diff
@@ -2,20 +2,20 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file function_requirements/zoom-downloader.txt function_requirements/zoom-downloader.in
+#    pip-compile --output-file=function_requirements/zoom-downloader.txt function_requirements/zoom-downloader.in
```
nataliemona, line 5: pip-compile added the "="

functions/zoom-downloader.py
```diff
@@ -119,17 +127,47 @@ def zoom_series_id(self):
...
+            e = datetime.strptime(self.end_time, TIMESTAMP_FORMAT)
+            delta = e - s
+            duration_in_mins = delta.seconds // 60
+        else:
+            # Default behavior
+            # do not filter out recording if duration cannot be found
+            duration_in_mins = MINIMUM_DURATION
+
+        return duration_in_mins < MINIMUM_DURATION
+
```
nataliemona, line 194: This was the original purpose of this branch. Filter out extremely short videos since we've had 2 tickets about sub-1-minute videos getting through to opencast this semester.

functions/zoom-downloader.py
```diff
@@ -152,25 +190,32 @@ def series_id_from_schedule(self):
...
+                    'F': 'Fridays'}
         if zoom_time.weekday() > 4:
             logger.debug("Meeting occurred on a weekend.")
             return None
-        elif weekdays[zoom_time.weekday()] not in schedule['Days']:
-            logger.debug("No opencast recording scheduled for this day of the week.")
+        letter = list(weekdays.keys())[zoom_time.weekday()]
+        if letter not in schedule['Days']:
+            logger.debug("No opencast recording scheduled on {}."
+                         .format(weekdays[letter]))
```
nataliemona, line 234: This doesn't  matter, I just wanted to print out the day of the week so the log reader doesn't have to go looking for it.

functions/zoom-downloader.py
```diff
@@ -152,25 +190,32 @@ def series_id_from_schedule(self):
...
             return None
 
-        threshold_minutes = 30
-        for time in schedule['Time']:
-            scheduled_time = datetime.strptime(time, '%H:%M')
+        for t in schedule['Time']:
```
nataliemona, line 237: I didn't want to shadow the name of the module "time"

template.yml
```diff
@@ -69,7 +69,7 @@ Resources:
       Environment:
         Type: "LINUX_CONTAINER"
         ComputeType: "BUILD_GENERAL1_SMALL"
-        Image: "aws/codebuild/eb-python-3.4-amazonlinux-64:2.1.6"
+        Image: "aws/codebuild/amazonlinux2-x86_64-standard:1.0"
```
nataliemona, line 72: Codebuild has an image with python 3.7 now!

functions/zoom-downloader.py
```diff
@@ -27,6 +27,14 @@
 CLASS_SCHEDULE_TABLE = env("CLASS_SCHEDULE_TABLE")
 LOCAL_TIME_ZONE = env("LOCAL_TIME_ZONE")
 
+# Recordings that happen within BUFFER_MINUTES a courses schedule
+# start time will be captured
+BUFFER_MINUTES = 30
+# Ignore recordings that are less than MIN_DURATION (in minutes)
+MINIMUM_DURATION = 2
```
lbjay, line 35: I trust that this will catch the majority of "ooops" videos? I mean, a 3 minute recording also seems like probably a mistake too.

functions/zoom-downloader.py
```diff
@@ -119,17 +127,55 @@ def zoom_series_id(self):
...
+                    format = "%H:%M:%S"
+                else:
+                    return False
+                duration_in_mins = time.strptime(self.duration, format).tm_min
+                return duration_in_mins < MINIMUM_DURATION
+            elif self.start_time and self.end_time:
+                s = datetime.strptime(self.start_time, TIMESTAMP_FORMAT)
+                e = datetime.strptime(self.end_time, TIMESTAMP_FORMAT)
+                delta = e - s
+                duration_in_mins = delta.seconds // 60
```
lbjay: TIL about he `//` operator. :)

functions/zoom-downloader.py
```diff
@@ -119,17 +127,55 @@ def zoom_series_id(self):
...
+    @property
+    def shorter_than_minimum_duration(self):
+        """
+        Interpret duration of meeting and determine whether duration is
+        shorter than the MINIMUM_DURATION in minutes.
+        """
+
+        try:
+            if self.duration:
+                if len(self.duration) == len("00:00"):
```
lbjay: this is a a little weird but fine. i've also used a try/catch loop when i'm not sure what format is going to work on an incoming string.

functions/zoom-webhook.py
```diff
@@ -75,7 +76,9 @@ def handler(event, context):
             "Handling not implemented for status '{}'".format(payload['status'])
         )
 
-    now = datetime.strftime(timezone(LOCAL_TIME_ZONE).localize(datetime.today()), '%Y-%m-%dT%H:%M:%SZ')
+    now = datetime.strftime(
```
lbjay, line 79: some of these reformattings suggest to me that you've set up pylint in atom :)

lbjay (approved):
:lgtm:

functions/zoom-downloader.py
nataliemona: True. We could probably increase it, maybe to 4. But looking at cloudwatch dashboard, most recordings <5 minutes are <2 minutes long.

functions/zoom-downloader.py
nataliemona: Floor division!

functions/zoom-downloader.py
nataliemona: I don't like it so I rewrote it. Changed the duration @property function so that it standardizes the duration format. Now shorter_than_minimum_duration can be much simpler. Also added tests.

PR #64, opened 2020-01-21
Add weekend scheduling and unit tests

nataliemona:


functions/zoom-downloader.py
```diff
@@ -214,24 +215,37 @@ def series_id_from_schedule(self):
         )
 
         if 'Item' not in r:
-            return None
+            self._class_schedule = None
         else:
             schedule = r['Item']
             logger.info(schedule)
+            self._class_schedule = schedule
```
lbjay: Is the idea to cache this value for the next time the property is read? if so, you're not actually looking for it anywhere.

functions/zoom-downloader.py
```diff
@@ -214,24 +215,37 @@ def series_id_from_schedule(self):
...
+
+        zoom_time = self.created.astimezone(timezone(LOCAL_TIME_ZONE))
+        days = {"M": "Mondays",
+                "T": "Tuesdays",
+                "W": "Wednesdays",
+                "R": "Thursdays",
+                "F": "Fridays",
+                "Sa": "Saturdays",
+                "Sn": "Sundays"}
+        day_code = list(days.keys())[zoom_time.weekday()]
```
lbjay, line 246: unless `days` is an `OrderedDict` I don't think you should rely on getting back the dict keys in a certain order. Maybe make `days` a list of tuples?

```
days = [("M", "Mondays"), ("T", "Tuesdays"), ("W", "Wednesdays"), ...]
day_code = [x[0] for x in days][zoom_time.weekday()]
```

tasks.py
```diff
@@ -230,16 +231,24 @@ def list_recordings(ctx, date=str(datetime.date.today())):
         print("Done!")
 
 
-@task(help={'uuid': 'meeting instance uuid', 'host-id': 'meeting host id'})
-def exec_pipeline(ctx, uuid, host_id, status=None, webhook_version=2):
+@task(help={'uuid': 'meeting instance uuid', 'host-id': 'meeting host id',
+            'ignore_schedule': ('do opencast series id lookup but ignore if '
```
lbjay, line 235: probably fine, but just thinking that anyone using these tasks without intimate knowledge of the code isn't going to know what this means.

tasks.py
```diff
@@ -1176,8 +1185,16 @@ def __schedule_json_to_dynamo(ctx, json_name):
 
     classes = json.load(file)
 
-    for item in classes.values():
-        table.put_item(Item=item)
+    try:
```
lbjay: This is fine too, but got me thinking it would be nice if we had a set of python lint rules that we re-used across projects (ala eslint). I would want it to complain about using `file = open(blah blah)` + `file.close()` vs a `with` statement that automatically closes the file handle.

tests/test_downloader.py
```diff
@@ -221,20 +225,95 @@ def test_duration(mocker):
...
 
+def local_hour(utc_timestamp):
+    utc = datetime.strptime(
+        utc_timestamp, TIMESTAMP_FORMAT) \
+        .replace(tzinfo=timezone('UTC'))
+    hour = utc.astimezone(timezone(LOCAL_TIME_ZONE)).hour
+    return "{:02}:00".format(hour)
+
+
+def test_series_id_from_schedule(mocker):
```
lbjay, line 250: nice :+1:

functions/zoom-downloader.py
nataliemona: You're right. There's no reason for me to do this. I was just messing around trying to get the unit test to work.

tasks.py
nataliemona: We will work on more intuitive ways to interact with zoom ingester and better documentation.

tasks.py
nataliemona: Yes, definitely.

PR #68, opened 2020-02-11
fixed duration filtering bug

nataliemona:


functions/zoom-downloader.py
```diff
@@ -188,10 +189,12 @@ def shorter_than_minimum_duration(self):
         shorter than the MINIMUM_DURATION in minutes.
         """
         if self.duration:
```
nataliemona, line 191: here's the bug, I thought tm_min would give total minutes

tests/test_downloader.py
```diff
@@ -228,6 +237,7 @@ def test_duration(mocker):
 def test_shorter_than_minimum_duration():
     cases = [
         ({"duration": "02:45:01"}, False),
```
nataliemona, line 239: and this test case had tm_min = 45 so it missed a corner case where tm_min = 1 or tm_min = 0

PR #69, opened 2020-02-28
Collect recording data from webhook

nataliemona:


function_requirements/zoom-webhook.txt
```diff
@@ -14,6 +14,7 @@ idna==2.8                 # via requests
 jmespath==0.9.4           # via boto3, botocore
 pyjwt==1.7.1
 python-dateutil==2.8.0    # via botocore
+python-dotenv==0.10.5
```
lbjay: is this a different version because of some other library in the requirements?

functions/common.py
```diff
@@ -1,14 +1,17 @@
...
 from os import getenv as env
 
-ZOOM_API_KEY = env('ZOOM_API_KEY')
-ZOOM_API_SECRET = env('ZOOM_API_SECRET')
 LOG_LEVEL = env('DEBUG') and 'DEBUG' or 'INFO'
 BOTO_LOG_LEVEL = env('BOTO_DEBUG') and 'DEBUG' or 'INFO'
+ZOOM_API_BASE_URL = "https://api.zoom.us/v2/"
+ZOOM_API_KEY = env("ZOOM_API_KEY")
+ZOOM_API_SECRET = env("ZOOM_API_SECRET")
+TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
```
lbjay, line 20: This value seems to be set in the other functions as well, but from an env var? Also it's not being used in `common.py`.

lbjay: oh, nevermind. it's the tests that get it from env. i see that the other functions are importing from common, so nevermind.

functions/common.py
```diff
@@ -44,9 +47,32 @@ def wrapped_func(event, context):
...
+            raise Exception(
+                "Call to ZoomAPIRequests.get "
+                "missing required param 'path'"
+            )
+
+        url = "{}{}".format(self.base_url, endpoint)
+        headers = {
+            "Authorization": "Bearer {}"
+            .format(self.gen_token(seconds_valid).decode())}
+        r = requests.get(url, headers=headers)
```
lbjay: This doesn't account for any timeout or connection errors. The `raise_for_status` method will only trigger if the api returns a non-200 response.

functions/common.py
```diff
@@ -44,9 +47,32 @@ def wrapped_func(event, context):
     return wrapped_func
 
 
-def gen_token(key=ZOOM_API_KEY, secret=ZOOM_API_SECRET, seconds_valid=60):
-    header = {"alg": "HS256", "typ": "JWT"}
-    payload = {"iss": key, "exp": int(time.time() + seconds_valid)}
-    return jwt.encode(payload, secret, headers=header)
+class ZoomApiRequest:
```
lbjay: It's nice that this is a class now, but the way you're using it is a little odd. Most cases, the point of making something like this a class is so that you can instantiate an object and reuse it. But now that there's so few actual calls to the api (getting the host name, getting the admin token), and in each case you're constructing a new object for each call, it doesn't look like you're getting any value out of the OOP.

functions/zoom-downloader.py
```diff
@@ -1,240 +1,215 @@
...
-        download_message = messages[0]
-        logger.info({'queue message': download_message})
-    except IndexError:
-        logger.info("No download queue messages available")
-        return
+        if not download_message.body:
+            logger.info("No messages available in downloads queue.")
+            return
+        if download_message.opencast_series_id:
+            break
```
lbjay, line 66: what happens to the sqs message here? does it not also need to be deleted?

functions/zoom-downloader.py
```diff
@@ -1,240 +1,215 @@
...
+            VisibilityTimeout=700
+        )
+        if (len(messages) == 0):
+            self.body = None
+            return None
 
-    @property
-    def uuid(self):
-        return self.data['uuid']
+        return messages[0]
```
lbjay: I think I'd like to see the message retrieval still take place outside of the class. Putting responsibilities like this in a constructor can sometimes work, but can make for confusing error handling if e.g. we discover later on there's exceptions that need to be handled. Can `message` be a thing that gets passed into the constructor?

functions/zoom-downloader.py
```diff
@@ -1,240 +1,215 @@
...
+    for _ in range(int(DOWNLOAD_MESSAGES_PER_INVOCATION)):
+        download_message = Download(
+            sqs, ignore_schedule, override_series_id
         )
-        download_message = messages[0]
-        logger.info({'queue message': download_message})
-    except IndexError:
-        logger.info("No download queue messages available")
-        return
+        if not download_message.body:
```
lbjay: this is kind of an awkward way to ascertain there's nothing in the queue. see my comment about the `Download` constructor below.

functions/zoom-downloader.py
```diff
@@ -1,240 +1,215 @@
...
+        return self._recording_files
 
     @property
-    def start_time(self):
-        if 'start_time' in self.data:
-            return self.data['start_time']
-        else:
-            return None
+    def uuid(self):
+        return self._uuid
```
lbjay: fwiw, i think it's not necessary to use these intermediary properties. i.e. you could just `return self.body["uuid"]` here.

functions/common.py
nataliemona: Does that make sense? I thought it would be better to have it defined in one place.

function_requirements/zoom-webhook.txt
nataliemona: This was generated by pip-compile so... not sure?

functions/common.py
nataliemona: The requests module will raise requests.ConnectionTimeout or requests.ConnectionError, so I think this is ok?

functions/zoom-downloader.py
nataliemona: No, because in this case the message has matched with the schedule so you will continue to process the message. The message will be deleted after sending a message to the upload queue or deadletter queue.

lbjay:
Something I noticed today while creating a new test stack is that `example.env` has `ZOOM_ADMIN_EMAIL` but the actual env value being used is `ZOOM_ADMIN_ID`.

**update** oops, nevermind. I see how it works now with `tasks.py` fetching the admin id using the email value.

lbjay:
Ahh, but `DOWNLOAD_MESSAGES_PER_INVOCATION` does need to be added to `example.env`

lbjay (approved):

PR #72, opened 2020-03-16
On-demand ingest

lbjay:
* new on-demand api endpoint + function
* new function for retrieving opencast operation counts
* uploader now checks for number of existing track uploads and aborts if > `OC_TRACK_UPLOAD_MAX`
* used `pip-compile` to update all dependencies to latest compatible versions
* new task to generate an api Resource Policy for applying access control to on-demand endpoint

Things still to do maybe:
* `OC_TRACK_UPLOAD_MAX` should be configurable
* downloader/uploader exec event schedule should be configurable


functions/opencast-op-counts.py
```diff
@@ -0,0 +1,50 @@
+import mysql.connector
+from os import getenv as env
+import urllib.parse as urlparse
+
+import logging
+from common import setup_logging
+logger = logging.getLogger()
+
+OPENCAST_DB_URL = env('OPENCAST_DB_URL');
```
nataliemona: Super low importance but I'm trying to use double quotes consistently instead of single quotes.

functions/zoom-on-demand.py
```diff
@@ -0,0 +1,110 @@
...
+    if uuid.startswith("https"):
+        # it's a link; parse the uuid from the query string
+        parsed_url = urlparse(uuid)
+        query_params = parse_qs(parsed_url.query)
+        if "meeting_id" not in query_params:
+            return resp(400, "Bad url: missing 'meeting_id' param.")
+        uuid = query_params["meeting_id"][0]
+
+    try:
+        zoom_endpoint = "/meetings/{}/recordings".format(uuid)
```
nataliemona, line 68: Just to note. This is a good endpoint in terms of rate limits. I believe it falls into the "light" category and I believe that means the limit is 60/second.

functions/zoom-on-demand.py
```diff
@@ -0,0 +1,110 @@
...
+        r = zoom_api_request(zoom_endpoint)
+        recording_data = r.json()
+    except requests.HTTPError as e:
+        if e.response.status_code == 404:
+            return resp(404, "No zoom recording with id '{}'"
+                        .format(uuid))
+        else:
+            return resp(500, "Something went wrong querying the zoom api: {}"
+                        .format(str(e)))
+
```
nataliemona, line 79: webhook lambda assumes all recordings are completed
so you might want to add some checks here?

This is from exec_webhook for example:

```
for file in data["recording_files"]:
        if "status" in file and file["status"] != "completed":
            raise Exception("Not all recordings have completed processing.")
        if "id" not in file and file["file_type"].lower() != "mp4":
            data["recording_files"].remove(file)
```

Or alternatively, webhook lambda should not assume that everything already has "status" of "completed"

functions/zoom-uploader.py
```diff
@@ -66,6 +74,16 @@ def sqs_resource():
...
+    current_uploads = get_current_upload_count()
+    if current_uploads is None:
+        logger.info("Unable to determine number of existing upload ops")
+        return
+    elif current_uploads >= OC_TRACK_UPLOAD_MAX:
+        logger.info("Too many current track uploads: {}".format(current_uploads))
+        return
+    else:
+        logger.info("Opencast upload count looks good: {}".format(current_uploads))
+
```
nataliemona, line 81: Looks good!

nataliemona (changes requested):
I also want the tests to pass. I don't mind fixing them myself tomorrow. It would probably be pretty quick for me since I just wrote them all and they are fresh in my head.

functions/opencast-op-counts.py
lbjay: :+1: 

I started to do `tasks.py` too but it was too much. Would blow up the diff and potentially introduce bugs in stuff that we wouldn't find until too late.

functions/zoom-on-demand.py
lbjay: :+1:

functions/zoom-on-demand.py
lbjay: :+1: 

Now checking that all recording files have status "completed" and returning a 503 if not.

lbjay:
Still didn't get to the failing tests but I think I addressed the other comments. Also made a few more things configurable. Soooo many template params / env vars...

nataliemona (approved):

PR #77, opened 2020-03-19
check for empty or missing recording files

lbjay:
* on-demand function checks if there's any recording files at all;
  if not returns a 503
* webhook now validates that any of the files are mp4 files;
  if not returns a 400
* rewrote a bunch of `test_webhook.py`

### DIFF
--- a/functions/zoom-webhook.py
+++ b/functions/zoom-webhook.py
@@ -130,6 +130,12 @@ def validate_payload(payload):
                     .format(field, obj.keys()))
 
         files = obj["recording_files"]
+
+        # make sure there's some mp4 files in here somewhere
+        mp4_files = any(x["file_type"].lower() == "mp4" for x in files)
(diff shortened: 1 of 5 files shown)


functions/zoom-webhook.py
```diff
@@ -130,6 +130,9 @@ def validate_payload(payload):
                     .format(field, obj.keys()))
 
         files = obj["recording_files"]
+        if not len(files):
+            raise BadWebhookData("No recording files in request payload")
+
```
nataliemona, line 138: Move this check to after the files are filtered by file type.

nataliemona (changes requested):

functions/zoom-webhook.py
lbjay: Now it's only confirming that any files are mp4

nataliemona (approved):

PR #78, opened 2020-03-19
Log groups and metric filters

nataliemona:


template.yml
```diff
@@ -827,6 +873,26 @@ Resources:
         MetricNamespace: !Sub "${AWS::StackName}-ZoomIngesterLogMetrics"
         MetricName: "MinutesInPipeline"
 
+  StartOperationMetricFilter:
+    Type: AWS::Logs::MetricFilter
+    Properties:
+      LogGroupName: !Ref OpencastOpCountsLogGroup
+      FilterPattern: "{$.message.operation_counts.START_OPERATION > 0}"
```
lbjay: these `START_OPERATION` / `START_WORKFLOW` operations aren't really relevant to anything

lbjay: I need to think about what we'd actually want to track here (if anything). Let's just skip these for now.

lbjay (changes requested):

lbjay (approved):

lbjay:
I had to revert this merge in master. Initially it was missing some resource renames, so I added an additional commit to fix that, but then running `stack.update` failed repeatedly because it couldn't replace some resources. Need to rework and test more thoroughly. The reflog entry for my follow-up commit is a1ff00a. Not sure if that made it into github. We can look at it and piece together next week.

PR #82, opened 2020-03-23
initial stab at fas import from csv

lbjay:


tasks.py
```diff
@@ -678,6 +677,73 @@ def get_day_of_week_code(day):
...
+        zoom_series_id = urlparse(row["zoom_link"]).path.split("/")[-1]
+        schedule_data.setdefault(zoom_series_id, {})
+        schedule_data[zoom_series_id]["zoom_series_id"] = zoom_series_id
+
+        opencast_series_id = urlparse(row["opencast_series_link"]) \
+            .fragment.replace("/", "")
+        schedule_data[zoom_series_id]["opencast_series_id"] = opencast_series_id
+
+        topic = "{} - {}".format(row["course_code"], row["title"])
+        schedule_data[zoom_series_id]["opencast_topic"] = topic
```
nataliemona: So, actually, this was "opencast_subject" in all my code.
We can change it to "opencast_topic" if you want.
Then I will update the spring schedule as well. It's confusing to try to read through the dynamo table with two columns for the same data. (Current situation in prod)

tasks.py
```diff
@@ -678,6 +677,73 @@ def get_day_of_week_code(day):
...
+            datetime.strftime(time_object, "%H:%M"),
+            (time_object + timedelta(minutes=30)).strftime("%H:%M"),
+            (time_object + timedelta(hours=1)).strftime("%H:%M")
+        ])
+
+    for id, item in schedule_data.items():
+        item["Days"] = list(item["Days"])
+        item["Time"] = list(item["Time"])
+
+    __schedule_json_to_dynamo(ctx, schedule_data=schedule_data)
```
nataliemona, line 750: This (__schedule_json_to_dynamo) prints out "schedule updated to" which is kind of useless when the schedule is this long and it hides the error messages printed out by your new code (i.e. "Invalid zoom link value for...")
Perhaps immediately, we should just remove that "schedule updated to" printout. What would be better would be to print out some kind of a diff.

tasks.py
```diff
@@ -678,6 +677,73 @@ def get_day_of_week_code(day):
...
+            raise Exit("Got bad day value: {}".format(row["day"]))
+
+        schedule_data[zoom_series_id].setdefault("Days", set())
+        schedule_data[zoom_series_id]["Days"].add(day)
+
+        schedule_data[zoom_series_id].setdefault("Time", set())
+        time_object = datetime.strptime(row["start"], "%H:%M")
+        schedule_data[zoom_series_id]["Time"].update([
+            datetime.strftime(time_object, "%H:%M"),
+            (time_object + timedelta(minutes=30)).strftime("%H:%M"),
```
nataliemona, line 742: In this spreadsheet we actually have the meeting duration so we could have these time points cover the duration of the meeting (and not extend past it). But this is lower priority. And maybe we don't want to rely on having the duration of the meeting for the time being if we may not always have it and it may not always be in every spreadsheet.

tasks.py
```diff
@@ -678,6 +677,73 @@ def get_day_of_week_code(day):
...
+        "producer",
+        "sit_in"
+    )
+    day_of_week_map = {
+        "Monday": "M",
+        "Tuesday": "T",
+        "Wednesday": "W",
+        "Thursday": "R",
+        "Friday": "F"
+    }
```
nataliemona, line 706: This excludes Saturday/Sunday. Will we ever have any Saturday/Sunday meetings?

tasks.py
```diff
@@ -974,7 +1041,7 @@ def __create_or_update(ctx, op):
                 getenv("ZOOM_API_KEY"),
                 getenv("ZOOM_API_SECRET"),
                 zoom_admin_id(),
-                "https://{}".format(oc_admin_host),
+                "http://{}".format(oc_admin_host),
```
nataliemona, line 1050: Why is this not HTTPS?

nataliemona (changes requested):

tasks.py
lbjay: man i coulda sworn i saw "_topic" somewhere. no, i'll change it to subject. i'll do a double update of the schedule (both sources) once this is merged and hopefully it'll get rid of all the items with the wrong column.

tasks.py
lbjay: yep, diff before updating would be even better. I also would like to look into a task arg that does a dry run and another that creates a backup of the table.

tasks.py
lbjay: i think it's fine how it is

tasks.py
lbjay: Because we've never reworked the opencast admin node to use https/ssl. :(
I didn't even think about it when I first typed "https" there, and then was puzzled when it was causing an error. Then I remembered...

We protect it in all sorts of other ways, but it's really bad that we haven't done this basic step.

PR #83, opened 2020-03-25
multi file uploads

lbjay:


lbjay:
Here's the context on how it chooses views:
* on the Opencast side we will now have an edit UI that allows producers to pick 2 out of 3 possible views that we upload
* paella can show only 2 streams at the moment (more coming in a future dev phase)
* the zoom-provided combined views, e.g. "shared_screen_with_speaker_view" are just pre-made versions of what paella can do when it has the individual streams
* therefore, we always prefer the individual views if present. It looks for any of those first, and if any are present that's what it uploads.
* if none of the individual views are available then it will go with one of the combined views

Q: Why would the zoom recording have the combined view streams but not the individual ones? I don't actually know. Is it maybe a host with incorrect recording settings? I'm seeing these as recently as yesterday. Most seem to be sections but there was at least one lecture.

Q: If we get only "primary_speaker" and "shared_screen_with_speaker_view" wouldn't it be better to go with the combined view that zoom provides? I'm not sure about this one either. To me it seems like this case should never happen, because if there's a "shared_screen_with_speaker_view" that would imply there's a "shared_screen" view too, right?

lbjay:
If you haven't seen the recording options in zoom it looks like this (thanks, mona)
image

functions/zoom-uploader.py
```diff
@@ -288,19 +281,43 @@ def ingest(self):
...
-            url = self._generate_presigned_url(s3_filename)
-            params.extend([
-                ("flavor", (None, escape(ZOOM_OPENCAST_FLAVOR))),
-                ("mediaUri", (None, url))
-            ])
+        file_params = self._generate_file_params()
+        if not len(file_params):
+            raise Exception("No recording file view types that we want")
+        else:
+            params.extend(file_params)
```
karendolan: @lbjay, the "multipart/chunked+source" is a minimum upload requirement (i.e. presenter). Is it ever possible that there would be a "shared_screen" and "gallery_view" but no "active_speaker"? 
If so, is it possible to fallback one of the  "shared_screen" or "gallery_view" to be flavor "multipart/chunked+source" on the upload to protect from errors?

functions/zoom-uploader.py
```diff
@@ -288,19 +281,43 @@ def ingest(self):
...
+        def make_param_pair(view_type, flavor):
+            if view_type not in self.s3_filenames:
+                return []
+            for s3_file in self.s3_filenames[view_type]:
+                logger.info("Found '{}' view: {}".format(view_type, s3_file))
+                s3_url = self._generate_presigned_url(s3_file)
+                return [
+                    ("flavor", (None, escape(flavor))),
+                    ("mediaUri", (None, s3_url))
+                ]
```
karendolan: Per @rute-santos , the order matters. If there is no "active_speaker", then "shared_screen"  should be assigned flavor ""multipart/chunked+source", and gallery to "presentation/chunked+source". If only gallery, it should be assigned "multipart/chunked+source". i.e. this looks like a straight mapping loop.

functions/zoom-uploader.py
nataliemona: _Is it ever possible that there would be a "shared_screen" and "gallery_view" but no "active_speaker"?_ -> Yes

rute-santos: Yes, Opencast requires at least the "presenter" file so, if there's only shared_screen and gallery, one of them needs to be uploaded as the multipart/chunked+source.
Also: If there's no "presentation", the gallery needs to be sent as the presentation/chunked+source. This is because the workflow cannot test all the possible combinations so if there's 1 video, it needs to be presenter; if there are 2, they need to be presenter & presentation.

lbjay:
OK, hopefully final update. There was some confusion on my part due to one of the flavors being called "gallery" when really it should just be "other" since it can end up being anything. Rute is fixing on her side as well.

functions/zoom-uploader.py
```diff
@@ -310,3 +394,28 @@ def _generate_presigned_url(self, s3_filename):
...
+        for primary_view, secondary_views in self.VIEW_PRIORITES.items():
+            if not self._has_view(primary_view) or self._has_presenter():
+                continue
+            self._add_presenter(primary_view)
+
+            for view in secondary_views:
+                if not self._has_view(view):
+                    continue
+                self._add_secondary(view)
+
```
nataliemona, line 424: The loop here continues even after the presenter or secondary is found. Meaning it actually doesn't select in priority order.

Also can you add logging for when each view is selected?

functions/zoom-uploader.py
```diff
@@ -288,19 +274,117 @@ def ingest(self):
...
+    # if we don't have either of the above then there can only be three
+    # possible views left and we just try them all in this order
+    FALLBACK_PRIORITIES = [
+        "shared_screen",
+        "shared_screen_with_gallery_view",
+        "gallery_view",
+    ]
+
+    MAX_FILES = 3
```
nataliemona: Do you need this or can you use len(FLAVORS)? If you keep it, can you rename it "MAX_VIEWS" and num_files "num_views" since you can still have multiple files per view (that will be concatenated), right?

The bigger problem here is that we aren't considering multiple tracks.

What if a "preferred" view exists but it doesn't span the duration of the class?

functions/zoom-uploader.py
```diff
@@ -288,19 +274,117 @@ def ingest(self):
             ("spatial", (None, "Zoom"))
         ]
 
-        for s3_filename in video_files:
-            url = self._generate_presigned_url(s3_filename)
-            params.extend([
-                ("flavor", (None, escape(ZOOM_OPENCAST_FLAVOR))),
-                ("mediaUri", (None, url))
-            ])
+        fpg = FileParamGenerator()
```
nataliemona: Should be
`FileParamGenerator(self.s3_filenames)`
and 
fpg.generate(self.s3_filenames) should just be `fpg.generate()`

functions/zoom-uploader.py
```diff
@@ -310,3 +394,28 @@ def _generate_presigned_url(self, s3_filename):
...
+            if not self._has_view(primary_view) or self._has_presenter():
+                continue
+            self._add_presenter(primary_view)
+
+            for view in secondary_views:
+                if not self._has_view(view):
+                    continue
+                self._add_secondary(view)
+
+        if self.num_files == 0:
```
nataliemona: Shouldn't this just be `if not self._has_presenter` ? At this point. If it already has a presenter then all of the fallback options would have already been checked for secondaries.

nataliemona (changes requested):

functions/zoom-uploader.py
lbjay: Logging added.

As for the loop, "secondary" can be either the presentation or "other" (gallery). The purpose of the loop is just to keep adding secondary views until we've got 3 total. Not sure what you mean by not selecting in priority order.

functions/zoom-downloader.py
```diff
@@ -140,22 +149,73 @@ def uuid(self):
...
+                        "file_count": len(track_set_files),
+                        "start": track_set_start,
+                        "end": track_set_end,
+                        "duration": duration_sec
+                    }
+                })
+
+                # skip any initial track sets < minimum duration
+                if not take_all and duration_sec < MINIMUM_DURATION * 60:
+                    logger.info("skipping track set {}".format(track_set_num))
```
nataliemona, line 208: Can we remove all segments < MINIMUM_DURATION? This problem could just as easily happen at the end or the middle of the recording (even if it isn't as often) - and this would cause you to lose the screen share. Tested with the labeled examples in the zoom ingester zoom account: https://harvard.zoom.us/recording

nataliemona (changes requested):

functions/zoom-downloader.py
lbjay: I get what you're saying. We can revisit this if turns out to be a problem. I'm not going to do it now because in the thing i sent around and producers signed off on i clearly stated we'd only remove them from the beginning and I'm not going to a) do another round of sign-off, or b) release something that doesn't do what we asked them to confirm (even if it's a pretty insignificant difference).

functions/zoom-uploader.py
lbjay: this was addressed as part of a general reworking

nataliemona (approved):

PR #85, opened 2020-03-31
ZIP-45: filenames not being sorted properly into multiple sequence sets

lbjay:
tbh I don't actually know what was wrong with the existing code. it was working very hard tho and resulting in files not getting sorted properly into correct sequence sets.

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -2,6 +2,7 @@ import boto3
 import json
 import requests
 from os import getenv as env
+from pathlib import Path
 from common import setup_logging, zoom_api_request, TIMESTAMP_FORMAT
 import subprocess
 from pytz import timezone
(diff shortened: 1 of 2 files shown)


functions/zoom-downloader.py
```diff
@@ -494,7 +495,7 @@ def stream(self):
 
     @property
     def file_extension(self):
-        return self.zoom_filename.split(".")[-1]
+        return splitext(self.zoom_filename)[-1]
```
nataliemona: This is creating a filename with two "."
Like so:
"577132153/1585344688/001-shared_screen_with_speaker_view..mp4"

functions/zoom-downloader.py
```diff
@@ -131,33 +132,33 @@ def host_name(self):
...
-                        ZoomFile(file_data, track_sequence)
-                    )
+                track_seq = start_times.index(file["recording_start"])
+                file["meeting_uuid"] = self.uuid
+                file["zoom_series_id"] = self.zoom_series_id
+                file["created_local"] = self._created_local
+                zoom_files.append(ZoomFile(file, track_seq))
+
+            self._recording_files = zoom_files
+
```
nataliemona, line 161: Would be nice to have:
logger.info(self._recording_files)
here

functions/zoom-downloader.py
```diff
@@ -131,33 +132,33 @@ def host_name(self):
...
             files = self.data["recording_files"]
-            start_times = sorted(
-                set([file["recording_start"] for file in files])
-            )
-            track_numbers = {
-                start_times[i]: i for i in range(len(start_times))
-            }
+            start_times = sorted(set(
+                [ file["recording_start"] for file in files ]
```
nataliemona: The spaces around the brackets are a linter error for me

nataliemona (changes requested):

functions/zoom-downloader.py
nataliemona: You can just remove the period in the s3_filename function

lbjay: oh wow good catch. i did not think it worked that way. i shouldn't be making unrelated changes like this anyway.

functions/zoom-downloader.py
lbjay: it's just going to print a bunch of `<ZoomFile...>` items, no?. You could make a separate PR to define a `__repr__` method for the ZoomFile class, and then add the log event.

functions/zoom-downloader.py
lbjay: how have I done this for so long without the pylint plugin installed?

PR #88, opened 2020-04-16
calculate estimated zoom processing time

nataliemona:
Calculate estimated Zoom processing time by using the start_time and duration from the webhook notification along with the time that the webhook notification was received. Do this only for notifications coming from Zoom (not for on demand ingests), and log this information so that it can be graphed in the dashboard.

### DIFF
--- a/functions/zoom-webhook.py
+++ b/functions/zoom-webhook.py
@@ -88,16 +88,11 @@ def handler(event, context):
         resp_callback = INGEST_EVENT_TYPES[zoom_event]
         return resp_callback(str(e))
 
-    sqs_message = construct_sqs_message(payload, context)
+    sqs_message = construct_sqs_message(payload, context, zoom_event)
     logger.info({"sqs_message": sqs_message})
 
-    if "delay_seconds" in payload:
-        logger.debug("Override default message delay.")
-        send_sqs_message(sqs_message, delay=payload["delay_seconds"])
-    elif zoom_event == "on.demand.ingest":
-        send_sqs_message(sqs_message, delay=0)
-    else:
-        send_sqs_message(sqs_message)
+    delay = 0 if zoom_event == "on.demand.ingest" else DEFAULT_MESSAGE_DELAY
+    send_sqs_message(sqs_message, delay=delay)
 
     return {
         "statusCode": 200,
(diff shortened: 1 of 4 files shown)


functions/zoom-webhook.py
```diff
@@ -86,16 +88,11 @@ def handler(event, context):
...
     logger.info({"sqs_message": sqs_message})
 
-    if "delay_seconds" in payload:
-        logger.debug("Override default message delay.")
-        send_sqs_message(sqs_message, delay=payload["delay_seconds"])
-    elif zoom_event == "on.demand.ingest":
-        send_sqs_message(sqs_message, delay=0)
-    else:
-        send_sqs_message(sqs_message)
+    delay = 0 if zoom_event == "on.demand.ingest" else DEFAULT_MESSAGE_DELAY
```
lbjay: eww, i don't like this one-liner if/else, but the existing code is also too much. I would prefer something like:

```
if zoom_event == "on.demand.ingest":
    delay = 0
else:
   delay = payload.get("delay_seconds", DEFAULT_MESSAGE_DELAY)
send_sqs_message(sqs_message, delay)
```

and then change the function signature of `send_sqs_message` to not use a default for the `delay` arg?

functions/zoom-webhook.py
```diff
@@ -201,11 +198,24 @@ def construct_sqs_message(payload, context):
     if "on_demand_series_id" in payload:
         sqs_message["on_demand_series_id"] = payload["on_demand_series_id"]
 
+    if zoom_event == "recording.completed":
+        zoom_processing_mins = estimated_processing_mins(
+            sqs_message["start_time"],
+            sqs_message["duration"]
+        )
+        sqs_message["zoom_processing_minutes"] = zoom_processing_mins
```
lbjay, line 209: i guess i'm not clear why the value has to get passed along in the sqs message. Is it only because of how it gets logged?

nataliemona: Yes, I'm pulling the metric from the sqs message log message. This isn't absolutely necessary but the alternative is having a separate log message just for this metric... and I'm trying to not clutter the log messages too much and it doesn't seem to be harmful to have more metadata in the sqs message.

Edit: Also, I forgot... it's also because I'm using other data in that log message (start time and duration, to make a graph). You can see if you go to the mona zoom ingester dashboard in aws dev.

functions/zoom-webhook.py
nataliemona: Updated

PR #93, opened 2020-06-05
multithreaded multipart transfer to s3

nataliemona:
Overall about a 40% reduction in runtime.

Tested with 3 sets of files:

5.8 GB set of files
w/o multithreading - 301s
with multithreading - 200s

7.4 GB set of files
w/o multithreading - 455s
with multithreading - 252s

16.2 GB set of files (largest I could find in our Zoom account)
w/o multithreading - timed out after 900s (15 minutes) max runtime limit for lambdas
with multithreading - 529s

Memory usage increased for all cases from about 100MB to 150-160MB.

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -9,6 +9,7 @@ from pytz import timezone
 from datetime import datetime
 from collections import OrderedDict
 import logging
+import concurrent.futures
 
 logger = logging.getLogger()
 
@@ -552,8 +553,17 @@ class ZoomFile:
     def file_extension(self):
         return Path(self.zoom_filename).suffix[1:]
 
+    def upload_part(self, part_number, chunk):
+        part = self.s3.upload_part(Body=chunk,
+                              Bucket=ZOOM_VIDEOS_BUCKET,
+                              Key=self.s3_filename,
+                              PartNumber=part_number,
+                              UploadId=self.upload_id)
+
+        return part
+
     def stream_file_to_s3(self):
-        s3 = boto3.client("s3")
+        self.s3 = boto3.client("s3")
 
         metadata = {
             "uuid": self.file_data["meeting_uuid"],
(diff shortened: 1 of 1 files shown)


functions/zoom-downloader.py
```diff
@@ -552,8 +553,17 @@ def stream(self):
...
+                              Bucket=ZOOM_VIDEOS_BUCKET,
+                              Key=self.s3_filename,
+                              PartNumber=part_number,
+                              UploadId=self.upload_id)
+
+        return part
+
     def stream_file_to_s3(self):
-        s3 = boto3.client("s3")
+        self.s3 = boto3.client("s3")
```
lbjay: i think just put this in the constructor

functions/zoom-downloader.py
```diff
@@ -565,41 +575,52 @@ def stream_file_to_s3(self):
...
              "metadata": metadata}
         )
-        part_info = {"Parts": []}
-        mpu = s3.create_multipart_upload(
+        parts = []
+        mpu = self.s3.create_multipart_upload(
                     Bucket=ZOOM_VIDEOS_BUCKET,
                     Key=self.s3_filename,
                     Metadata=metadata)
+        self.upload_id = mpu["UploadId"]
```
lbjay: since the object/class doesn't use this value anywhere else i wonder if it shouldn't just be another arg to `upload_part`?

functions/zoom-downloader.py
```diff
@@ -565,41 +575,52 @@ def stream_file_to_s3(self):
...
             print("Completed multipart upload of {}.".format(self.s3_filename))
         except Exception as e:
             logger.exception(
                 "Something went wrong with upload of {}:{}"
                 .format(self.s3_filename, e)
             )
-            s3.abort_multipart_upload(Bucket=ZOOM_VIDEOS_BUCKET,
-                                      Key=self.s3_filename,
-                                      UploadId=mpu["UploadId"])
+            self.s3.abort_multipart_upload(
```
lbjay, line 619: I wonder a little about if there are ways for the concurrent.futures stuff to go wrong that wouldn't end up here. Just thinking out loud, not asking for changes. I guess something could get hung and the whole lambda times out (which already can happen). Wasn't there something we had to do once related to half-finished multi-part bucket files?

lbjay (approved):
Looks great

functions/zoom-downloader.py
nataliemona: We have a deletion policy for parts uploaded and abandoned by a failed multipart upload. I think that's what you are thinking of?

PR #94, opened 2020-06-09
allow_multiple_ingests flag, invoke exec.on-demand-ingest task

nataliemona:
Add a field to the on demand ingest request (/ingest endpoint) that allows multiple ingests by generating a random mpid (in the uploader lambda) rather than a mpid based on the recording uuid.

Add a exec.on-demand-ingest task to conveniently invoke the /ingest endpoint.


lbjay (approved):
looks great

PR #95, opened 2020-06-10
fix ffprobe and extract file metadata

nataliemona:
Goal here was to log some useful file metadata (size and duration) in the uploader so we can have some metrics for what is ingested into opencast.
While working on that, I realized that we weren't calling the method that runs ffprobe.
Fixed ffprobe and added a lot of file metadata to the uploader message.

Tested with the 16.4 GB set of files to make sure we weren't taking a large hit in terms of runtime and the downloader lambda completed in 8 minutes 39 seconds so it seems ok.

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -341,12 +341,39 @@ class Download:
     @property
     def upload_message(self):
         if not hasattr(self, "self._upload_message"):
-            s3_filenames = {}
+            s3_files = {}
+            segment_durations = {}
             for file in self.recording_files:
-                if file.recording_type in s3_filenames:
-                    s3_filenames[file.recording_type].append(file.s3_filename)
+                segment = {
+                    "filename": file.s3_filename,
+                    "recording_start": file.file_data["recording_start"],
+                    "recording_end": file.file_data["recording_end"],
+                    "ffprobe_bytes": file.file_data["ffprobe_bytes"],
+                    "ffprobe_seconds": file.file_data["ffprobe_seconds"]
+                }
+                segment_durations[segment["recording_start"]] = segment["ffprobe_seconds"]
+                if file.recording_type in s3_files:
+                    s3_files[file.recording_type]["segments"].append(segment)
                 else:
-                    s3_filenames[file.recording_type] = [file.s3_filename]
+                    s3_files[file.recording_type] = {
+                        "segments": [segment]
+                    }
+
+            all_file_bytes = 0
+            all_file_seconds = 0
+            for view_data in s3_files.values():
+                view_data["view_bytes"] = sum(
+                    s["ffprobe_bytes"] for s in view_data["segments"]
+                )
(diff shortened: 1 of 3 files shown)


functions/zoom-downloader.py
```diff
@@ -624,7 +673,7 @@ def stream_file_to_s3(self):
             raise
 
         if self.file_extension == "mp4":
-            if not self.valid_mp4_file:
+            if not self.valid_mp4_file():
```
nataliemona, line 679: So it turns out because "valid_mp4_file" wasn't even being called here, we were not even running ffprobe!!!

functions/zoom-uploader.py
```diff
@@ -210,7 +210,12 @@ def workflow_definition_id(self):
 
     @property
     def s3_filenames(self):
-        return self.data["s3_filenames"]
+        if not hasattr(self, "_s3_filenames"):
+            self._s3_filenames = {}
+            for view, file in self.data["s3_files"].items():
+                self._s3_filenames[view] = [data["filename"] for data in file["segments"]]
+
+        return self._s3_filenames
```
nataliemona, line 239: It just seemed easiest, rather than mess with the uploader code, to give it the data structure it expects.

functions/zoom-downloader.py
```diff
@@ -341,12 +341,39 @@ def oc_series_found(self, ignore_schedule=False, override_series_id=None):
...
+                )
+                view_data["view_seconds"] = sum(
+                    s["ffprobe_seconds"] for s in view_data["segments"]
+                )
+                all_file_bytes += view_data["view_bytes"]
+                all_file_seconds += view_data["view_seconds"]
+
+            recording_seconds = sum(
+                duration for duration in segment_durations.values()
+            )
```
nataliemona, line 377: Calculating this `recording_seconds` number because it turns out the duration number we get from zoom seems to be wildy off in comparison to the ffprobe output.

functions/zoom-downloader.py
```diff
@@ -521,18 +552,36 @@ def s3_filename(self):
...
-        s3 = boto3.client("s3")
-        url = s3.generate_presigned_url(
+        url = self.s3.generate_presigned_url(
             "get_object",
             Params={"Bucket": ZOOM_VIDEOS_BUCKET, "Key": self.s3_filename}
         )
 
-        command = ["/var/task/ffprobe", "-of", "json", url]
-        if subprocess.call(command) == 1:
+        cmd = f"/var/task/ffprobe -of json -show_format {url}"
```
nataliemona, line 563: In addition to never calling `valid_mp4_file`. If it were to be called we would not have be getting any useful data out of ffprobe anyway!
This `-show_format` flag gets us some useful output.

functions/zoom-downloader.py
lbjay: whoops!

lbjay (approved):

PR #98, opened 2020-06-20
Bug fix: infinite recursion when recording is already ingested

nataliemona:
Bug fix. The mediapackage_id function called itself in order to print out the mediapackage_id when a recording was already ingested and allow_multiple_ingests was false.

Also added unit tests.


functions/zoom-uploader.py
```diff
@@ -181,8 +180,9 @@ def mediapackage_id(self):
                 else:
                     logger.warning(
                         "Episode with deterministic mediapackage id"
-                        f" {self.mediapackage_id} already ingested"
```
nataliemona, line 184: whoops this is recursion

PR #99, opened 2020-06-23
add a log message post ingest decision

nataliemona:


functions/zoom-uploader.py
```diff
@@ -94,26 +94,22 @@ def handler(event, context):
     upload_message = messages[0]
     logger.debug({
         "queue_message": {
-            "attributes": upload_message.attributes,
```
nataliemona, line 97: this is always null

functions/zoom-uploader.py
```diff
@@ -94,26 +94,22 @@ def handler(event, context):
...
             "body": upload_message.body
         }
     })
 
     try:
         upload_data = json.loads(upload_message.body)
-        logger.info({
```
nataliemona, line 104: move to after ingest decision

functions/zoom-uploader.py
```diff
@@ -142,8 +138,8 @@ def get_current_upload_count():
 
 def process_upload(upload_data):
     upload = Upload(upload_data)
-    upload.upload()
-    return upload.workflow_id
```
nataliemona, line 146: this was calling the function workflow_id multiple times

functions/zoom-uploader.py
```diff
@@ -239,10 +235,10 @@ def s3_filenames(self):
 
     @property
     def workflow_id(self):
-        if not hasattr(self, "workflow_xml"):
-            logger.warning("No workflow xml yet!")
-            return None
         if not hasattr(self, "_workflow_id"):
+            if not hasattr(self, "workflow_xml"):
```
nataliemona, line 239: workflow_xml should only be checked once

tasks.py
```diff
@@ -84,6 +84,9 @@ def production_failsafe(ctx):
     It is meant to be prepended to the execution of other tasks to force a confirmation
     when a task is being executed that could have an impact on a production stack
     """
+    if not STACK_NAME:
+        raise Exit("No STACK_NAME specified")
+
```
nataliemona, line 89: if prod identifier is an empty string and STACK_NAME is not defined, then the production failsafe will be triggered and then the user will get an error message that doesn't make it clear that the problem is that STACK_NAME is not defined

tasks.py
```diff
@@ -338,15 +341,15 @@ def exec_on_demand(ctx, uuid, oc_series_id=None, allow_multiple_ingests=False):
 @task(help={'uuid': 'meeting instance uuid',
             'ignore_schedule': ('ignore schedule, use default series if '
                                 'available'),
-            'on_demand_series_id': ('opencast series id to use regardless of '
+            'oc_series_id': ('opencast series id to use regardless of '
```
nataliemona, line 344: change to make exec.on-demand and exec.webhook/exec.pipeline consistent

tasks.py
```diff
@@ -1214,6 +1217,7 @@ def __publish_version(ctx, func, description):
 
 
 def __build_function(ctx, func, upload_to_s3=False):
+    print(f"Building {func} function")
```
nataliemona, line 1220: Add print statements that explain the steps and hide=1 to stop the terminal spam

functions/zoom-webhook.py
```diff
@@ -207,6 +207,10 @@ def construct_sqs_message(payload, context, zoom_event):
     if "on_demand_series_id" in payload:
         sqs_message["on_demand_series_id"] = payload["on_demand_series_id"]
 
+    # not used in downloader or uploader but useful for cloudwatch dashboard
```
nataliemona, line 210: I want to have the zoom processing times, and the size of the file available in one log message - it seems like zoom processing times are mostly impacted by file size so this is important.

lbjay (approved):

PR #100, opened 2020-06-25
For reference - CDK edits

nataliemona:


cdk/api.py
```diff
@@ -8,15 +8,16 @@ class ZipApi(core.Construct):
 
     def __init__(self, scope: core.Construct, id: str,
             webhook_function,
-            on_demand_function):
+            on_demand_function,
+            lambda_release_alias):
```
nataliemona, line 12: this env var was missing from stack props, and also not passed in here

cdk/api.py
```diff
@@ -8,15 +8,16 @@ class ZipApi(core.Construct):
...
         super().__init__(scope, id)
 
         stack_name = core.Stack.of(self).stack_name
 
         self.rest_api_name = f"{stack_name}-api"
         self.api = apigw.LambdaRestApi(
             self, "api",
-            handler=webhook_function,  # default handler
+            handler=webhook_function.function,  # default handler
             rest_api_name=self.rest_api_name,
```
nataliemona, line 21: here you're mixing your classes, for example `<class 'cdk.function.ZipWebhookFunction'>` with the cdk classes such as `<class 'aws_cdk.aws_lambda.Function'>`

cdk/api.py
```diff
@@ -44,7 +45,7 @@ def __init__(self, scope: core.Construct, id: str,
         )
 
         self.ingest_resource = self.api.root.add_resource("ingest")
-        on_demand_integration = apigw.LambdaIntegration(on_demand_function)
+        on_demand_integration = apigw.LambdaIntegration(on_demand_function.function)
         self.ingest_method = self.ingest_resource.add_method(
```
nataliemona, line 49: your class vs cdk class - there's a few of these

cdk/api.py
```diff
@@ -58,25 +59,22 @@ def __init__(self, scope: core.Construct, id: str,
...
+        for metric_name in ["4XX", "5XX"]:
+                alarm = cloudwatch.Alarm(self, f"ZoomWebhook{metric_name}Alarm",
                     metric=cloudwatch.Metric(
                         metric_name=metric_name,
                         namespace="AWS/ApiGateway",
                         dimensions={
                             "ApiName": self.rest_api_name,
-                            "Stage": "live",
-                            "Method": "POST",
-                            "Resource": resource.path,
```
nataliemona: you can just have one 4xx and 5xx alarm for the whole api... if you want to have them per resource... which actually is probably a good idea so we can stop getting annoying emails every time somebody puts the wrong link into the Zoom+ tool then the metric names have to be unique

cdk/queues.py
```diff
@@ -51,6 +51,7 @@ def add_monitoring(self, monitoring):
             metric=self.upload_queue.metric("ApproximateNumberOfMessagesVisible"),
             statistic="sum",
             threshold=40,
+            evaluation_periods=1,
             period=core.Duration.minutes(5),
```
nataliemona, line 55: evaluation_periods was missing and is required

cdk/api.py
```diff
@@ -58,25 +59,22 @@ def __init__(self, scope: core.Construct, id: str,
             ]
         )
 
-        on_demand_function.add_environment(
+        on_demand_function.function.add_environment(
             "WEBHOOK_ENDPOINT_URL",
-            (f"https://{self.api.rest_api_id}.execute-api.{self.region}"
-             f".amazonaws.com/{self.lambda_release_alias}/new_recording")
+            (f"https://{self.api.rest_api_id}.execute-api.{core.Stack.of(self).region}"
```
nataliemona, line 64: region of the stack not the api

cdk/api.py
```diff
@@ -58,16 +59,17 @@ def __init__(self, scope: core.Construct, id: str,
...
 
     def add_monitoring(self, monitoring):
 
-        for metric_name in ["4XXError", "5XXError"]:
+        for error_type in ["4XX", "5XX"]:
             for resource in [self.new_recording_resource, self.ingest_resource]:
+                metric_name = f"{resource.path.strip('/')}{error_type}Error"
                 alarm = cloudwatch.Alarm(self, f"{metric_name}Alarm",
                     metric=cloudwatch.Metric(
                         metric_name=metric_name,
```
nataliemona, line 75: In our current setup we have one set of 4xx and 5xx alarms for the whole api... I like the idea of having them per resource as you did here... but there is an error if they don't have unique names. Also, I think I would prefer to not have the 4xx alarm for the ingest endpoint since those error emails are just people putting the incorrect link into the Zoom+ tool.

cdk/function.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, scope: core.Construct, id: str,
         }
 
         if vpc_id is not None and security_group_id is not None:
-            opencast_vpc = ec2.Vpc.from_lookup(self, "OpencastVpc", vpc_id=vpc_id),
```
nataliemona: ¯\_(ツ)_/¯ python

cdk/api.py
lbjay: I started trying to do away with this setting and just hardcoding "live" everywhere. It's in some of the tasks and the codebuild stuff too, so I guess I still need to choose which direction to go on this. I feel like we've never figured out how to take advantage of having this as a variable. Do you want to make an argument for it?

cdk/api.py
lbjay: yeah i was doing that in a few other places too. Pycharm helps flag some of these kinds of things but misses a lot too. It kindof makes me miss the stronger typing of typescript. I know you can bring some of that typing into python with the jsii stuff, but not sure how far to take that. Seems un-pythonic. It's also possible there's a way to make pycharm smarter about it's analysis?

cdk/api.py
lbjay: haha yep.

cdk/api.py
lbjay: i decided to do them separately since we could. fixed the resource ids.

cdk/queues.py
lbjay: yep got it

cdk/api.py
lbjay: got it

cdk/api.py
lbjay: hmm, i agree we don't want an alert every time that happens, but now i'm wondering what kind of 4xx errors do we get for the webhook endpoint? also if I could stop trying to be such a HTTP purist would it make sense to have the on-demand function return something besides a 4xx in those cases of producer error? nah forget it. IIRC we'd have to make a change on the opencast side too. 

A better question is, is there a reason to get an alert for **any** 4xx for either endpiont?

PR #103, opened 2020-08-11
filter false starts based on ffprobe duration data

nataliemona:
Commentary on a better place to do the filtering welcome. It's a little difficult because we cannot filter before we transfer the files to S3 (since we need the ffprobe data) but after we transfer the files to S3 the files are organized by view not segment in order to make the filtering by view easy in the uploader.

### DIFF
--- a/functions/zoom-downloader.py
+++ b/functions/zoom-downloader.py
@@ -161,27 +161,28 @@ class Download:
             # all incoming files
             files = self.data["recording_files"]
 
-            # number of distinct start times is the count of track sets
-            track_set_start_times = sorted(set(
+            # number of distinct start times is the count of segements
+            segment_start_times = sorted(set(
                 [file["recording_start"] for file in files]
             ))
 
             # collect the recording files here as ZoomFile objs
             zoom_files = []
 
-            # track sets refers to the sets of recording files that
+            # segement refers to the sets of recording files that
             # are generated when a host stops and restarts a meeting
-            for track_set_num, start_time in enumerate(track_set_start_times):
+            # NOTE: segment numbering starts at 0 (zero)
+            for segment_num, start_time in enumerate(segment_start_times):
                 # all the files from this segment
-                track_set_files = [
+                segment_files = [
                     f for f in files if f["recording_start"] == start_time
                 ]
 
-                for file in track_set_files:
+                for file in segment_files:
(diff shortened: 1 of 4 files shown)


functions/zoom-downloader.py
```diff
@@ -169,48 +169,13 @@ def recording_files(self):
...
-                # skip any initial track sets < minimum duration
-                if not take_all and duration_sec < MINIMUM_DURATION * 60:
-                    logger.info("skipping track set {}".format(track_set_num))
-                    continue
-
-                # once past the minimum duration we take everything else
-                take_all = True
+                track_set_files = [
+                    f for f in files if f["recording_start"] == start_time
+                ]
```
lbjay: :+1:

functions/zoom-downloader.py
```diff
@@ -169,48 +169,13 @@ def recording_files(self):
...
-                             if f["recording_start"] == start_time]
-                # parse their start/end times
-                track_set_start = datetime.strptime(
-                    track_set_files[0]["recording_start"],
-                    TIMESTAMP_FORMAT
-                )
-                track_set_end = datetime.strptime(
-                    track_set_files[0]["recording_end"],
-                    TIMESTAMP_FORMAT
-                )
```
lbjay: I remember hating writing this code. RIP.

functions/zoom-uploader.py
```diff
@@ -228,8 +230,21 @@ def workflow_definition_id(self):
...
             self._s3_filenames = {}
-            for view, file in self.data["s3_files"].items():
-                self._s3_filenames[view] = [data["filename"] for data in file["segments"]]
+            for view, file_info in self.data["s3_files"].items():
+                segments = file_info["segments"]
+
+                # skip false starts
+                if segments[0]["ffprobe_seconds"] < MINIMUM_DURATION * 60:
+                    # first segment for each view might not be the overall
+                    # first segment, need to check
```
lbjay: Can you elaborate on this some more?

nataliemona: You might have.
2pm - 15s
- speaker

2:01pm - 1m
- speaker
- shared screen

2:15pm - 1 hour
- speaker
- shared screen

If you remove the first segment in each view that is <2 minutes then your segments will not line up.

PR #107, opened 2020-09-29
f/schedule update endpoint

nataliemona:


functions/schedule-update.py
```diff
@@ -0,0 +1,28 @@
...
+    TODO: Explain the purpose of this function.
+    """
+    sheet = GSheet(GSHEETS_DOC_ID, in_lambda=True)
+    sheet.import_to_dynamo(GSHEETS_SHEET_NAME)
+
+    return {
+        "statusCode": 200,
+        "headers": {},
+        "body": "Success"
+    }
```
nataliemona: missing new line - I'll fix my editor

functions/schedule-update.py
```diff
@@ -0,0 +1,28 @@
...
+GSHEETS_DOC_ID = env("GSHEETS_DOC_ID")
+GSHEETS_SHEET_NAME = env("GSHEETS_SHEET_NAME")
+
+logger = logging.getLogger()
+
+
+@setup_logging
+def handler(event, context):
+    """
+    TODO: Explain the purpose of this function.
```
nataliemona: remove this or fix this

functions/gsheets.py
```diff
@@ -0,0 +1,238 @@
...
+        Save credentials file to SSM parameter.
+        """
+        if not filename:
+            filename = self.default_filename
+        if not exists(filename):
+            raise Exception(f"Credentials file {filename} not found")
+
+        data = ""
+        with open(filename, "r") as f:
+            data = json.dumps(f.read())
```
lbjay: This doesn't seem right. I would expect a `json.load(f)` here.

cdk/stack.py
```diff
@@ -66,6 +71,24 @@ def __init__(self, scope: core.Construct, id: str,
...
+            lambda_code_bucket=lambda_code_bucket,
+            environment={
+                "CLASS_SCHEDULE_TABLE": schedule.table.table_name,
+                "GSHEETS_DOC_ID": gsheets_doc_id,
+                "GSHEETS_SHEET_NAME": gsheets_sheet_name,
+            }
+        )
+        schedule_update.function.add_to_role_policy(
+            iam.PolicyStatement(
+                actions=["ssm:GetParameter", "ssm:PutParameter"],
```
lbjay, line 85: does the function ever need to write to parameter store?

functions/gsheets.py
```diff
@@ -0,0 +1,238 @@
...
+from googleapiclient.discovery import build
+
+SCHEDULE_TABLE = env("CLASS_SCHEDULE_TABLE")
+
+logger = logging.getLogger()
+
+load_dotenv(join(dirname(__file__), '../.env'))
+
+
+class GSheetsAuth():
```
lbjay, line 25: It's maybe not worth redoing, but something to think about in the future re: putting stuff in classes. I'm not sure it's warranted in this situation. Usually you would do that if you wanted to have functions (methods) that shared some kind of state between calls. Here it seems like any given process is only going to loading, saving or deleting once.

functions/gsheets.py
```diff
@@ -0,0 +1,238 @@
...
+        r = self.ssm.put_parameter(
+            Name=self.ssm_path,
+            Description="Service account creds for gsheets authorization",
+            Value=json.loads(data),
+            Type="SecureString",
+            Overwrite=True,
+        )
+        logger.info(f"Saved token version {r['Version']}")
+
+    def delete_from_ssm(self):
```
lbjay, line 78: does this get run on stack.delete?

tasks.py
```diff
@@ -590,96 +601,18 @@ def get_day_of_week_code(day):
             raise Exit("Failed converting to dynamo item format: {}\n{}" \
                        .format(str(e), json.dumps(event, indent=2)))
 
-    __schedule_json_to_dynamo(ctx, schedule_data=schedule_data)
+    schedule_json_to_dynamo(
```
lbjay: this is calling `schedule_json_to_dyanmo` with an invalid arg set. do we still need this import from opencast task?

functions/gsheets.py
```diff
@@ -0,0 +1,238 @@
...
+        logger.debug(f"Successfully downloaded {sheet_name}.csv")
+        return file_path
+
+    def import_to_dynamo(self, sheet_name):
+        file_path = self._download_csv(sheet_name)
+
+        schedule_csv_to_dynamo(SCHEDULE_TABLE, file_path)
+
+
+def schedule_json_to_dynamo(schedule_table, json_file=None, schedule_data=None):
```
lbjay: not asking for changes; just thinking out loud. i wrote this code iirc and it feels a bit inelegant and possibly error prone. I was fine with that when it only ran locally and i could play with it and comment out stuff. Now that it's going to run in a lambda that has to be packaged, deployed, etc, I wonder if we're going to wish it was broken up a bit and more testable. IDK, what do you think?

lbjay (changes requested):

cdk/stack.py
nataliemona: No

functions/gsheets.py
nataliemona: Thanks :-)

functions/gsheets.py
nataliemona: No... maybe we should just remove this since credentials are being shared.

tasks.py
nataliemona: Good point. Let's keep the import form opencast code though until we're sure we won't be needing it anymore?

functions/gsheets.py
nataliemona: Added a unit test for the parsing code - should add more as well.

tasks.py
lbjay: That's fine. It's always in the git history though, don't forget.

PR #108, opened 2020-09-30
spreadsheet gscript that assembles and pushes the schedule csv

lbjay:
all this is is me adding the gscript + removing enough of the original google api auth stuff to get my stack to update. the `schedule-update.py` lambda just dumps the event to the logs (which is bad because of the zoom links in there, so i just went and deleted those log streams.


lbjay:
Closing this as i think we've settled on the correct google api impl.

PR #109, opened 2020-10-27
b/ZIP-56-deleted files race condition

nataliemona:
Fix for the case where a file is deleted between the time when the webhook notification is received from Zoom and when the files are downloaded by the downloader lambda. ZIP should skip over any dead links and only throw a PermanentDownloadError if none of the files are downloadable.

Also, add permissions in CDK to downloader and uploader lambda to access the deadletter queues.


lbjay (approved):

lbjay:
Remove the `ZOOM_ADMIN_ID = env("ZOOM_ADMIN_ID")` line at the top of the downloader function if it's not used anymore. And the `gen_admin_token` function too.

PR #111, opened 2020-11-19
F/zip 55 titles typenums

nataliemona:


functions/zoom-downloader.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, sqs, data):
         self.sqs = sqs
         self.data = data
         self.opencast_series_id = None
+        self.title = "Lecture"  # default title
```
nataliemona, line 134: Will be used for on demand ingest

functions/zoom-downloader.py
```diff
@@ -350,6 +357,7 @@ def upload_message(self):
                 "uuid": self.data["uuid"],
                 "zoom_series_id": self.data["zoom_series_id"],
                 "opencast_series_id": self.opencast_series_id,
+                "oc_title": self.title,
```
nataliemona, line 360: Lecture, Section, or something else

functions/gsheets.py
```diff
@@ -188,25 +190,31 @@ def lower_case_first_line(iter):
...
-
-        days_value = row["day"].strip()
+        # Add descriptions
+        schedule_data[zoom_series_id]["course_code"] = row["course code"]
 
+        # Construct events - list of {day, time, title}
+        schedule_data[zoom_series_id].setdefault("events", [])
+        title = row["type"].strip().title()
+        time_object = datetime.strptime(row["start"], "%H:%M")
+        time = datetime.strftime(time_object, "%H:%M")
```
lbjay: is the point here just validation? checking that the value goes back and forth?

functions/gsheets.py
```diff
@@ -221,20 +229,16 @@ def lower_case_first_line(iter):
...
+            schedule_data[zoom_series_id]["events"].append(
+                {"day": day, "time": time, "title": title}
+            )
 
     for item in schedule_data.values():
         # sort days for easier testing
-        item["Days"] = sorted(
-            list(item["Days"]), key=lambda x: valid_days.index(x[0])
+        item["events"] = sorted(
+            list(item["events"]), key=lambda x: valid_days.index(x["day"])
```
lbjay: trying to understand what the `list()` is doing here

functions/zoom-downloader.py
```diff
@@ -222,13 +223,13 @@ def _class_schedule(self):
...
-        if "Item" not in r:
-            return None
-        else:
+        if "Item" in r:
             schedule = r["Item"]
             # DynamoDB sometimes returns type decimal.Decimal
             schedule["opencast_series_id"] = str(schedule["opencast_series_id"])
             return schedule
+        else:
+            return None
```
lbjay: you made me google: https://stackoverflow.com/questions/4027586/is-it-ok-to-skip-return-none

personally, i prefer returning None early if some condition isn't met. Then, if I'm being mindful about it, i'd skip the `else:` entirely.  There's some airbnb eslint rule that actually complains when I do that in javascript. But whatever.

functions/zoom-downloader.py
```diff
@@ -254,26 +252,35 @@ def _series_id_from_schedule(self):
...
+        zoom_time = self._created_local
+        logger.info({"meeting creation time": zoom_time,
+                     "course schedule": schedule})
+        zoom_day_code = list(days.keys())[zoom_time.weekday()]
+
+        # events is a list of {title, day, time} dictionaries
+        for event in schedule["events"]:
+            # match day
+            if zoom_day_code != event["day"]:
+                continue
```
lbjay, line 265: Like airbnb eslint would complain about this `continue` here. I guess it wants you to just endessly nest `if` statements? IDK it really bugs me how wrong that is.

lbjay (approved):
looks great

functions/gsheets.py
nataliemona: Yes, rewriting this to make more sense as a validation check.

functions/gsheets.py
nataliemona: Was leftover from the previous parsing code. I think it's no longer necessary.

functions/zoom-downloader.py
nataliemona: Updated

PR #112, opened 2020-12-03
ignore minimum duration check if on demand ingest

nataliemona:


functions/zoom-downloader.py
```diff
@@ -72,7 +72,7 @@ def handler(event, context):
         # this is checking for ~total~ duration of the recording as reported
         # by zoom in the webhook payload data. There is a separate check later
         # for the duration potentially different sets of files
-        if dl.duration >= MINIMUM_DURATION:
+        if dl.duration >= MINIMUM_DURATION or "on_demand_series_id" in dl_data:
```
lbjay, line 75: series id is optional when the user submitting has opencast admin privs

lbjay (approved):

PR #113, opened 2021-01-05
T/apigee support

nataliemona:


functions/common.py
```diff
@@ -63,23 +64,27 @@ def gen_token(key, secret, seconds_valid=60):
...
-            )
+def zoom_api_request(endpoint, seconds_valid=60, ignore_failure=False, retries=3):
+    if not endpoint:
+        raise Exception("Call to zoom_api_request missing endpoint")
+
+    if not APIGEE_KEY and not (ZOOM_API_KEY and ZOOM_API_SECRET):
+        raise Exception(("Missing api credentials. "
+            "Must have APIGEE_KEY or ZOOM_API_KEY and ZOOM_API_SECRET"))
 
     url = "{}{}".format(ZOOM_API_BASE_URL, endpoint)
```
lbjay: I'd love to see this done in a way that avoids any issues/confusion about the base url having a dangling '/' or the `endpoint` value needing (or not) to be prefixed by '/'. OTOH, `urljoin` can be unintuitive and has bitten me before. I can't believe it's 2021 and there's no simple way to do this.

lbjay (commented):
I have concerns about the process of reverting back to the zoom api key/secret if for some reason we needed to bypass apigee. From the code-perspective the logic seems fine; if `APIGEE_KEY` is None then it'll fall back to the zoom key/secret, but what does that mean in terms of deploying that change? We would unset the value in the `.env` and then do a stack update? We're confident the env var wouldn't get an empty string (vs `None)?

PR #114, opened 2021-01-19
F/status tracking 2021

nataliemona:
Query the status endpoint with:

`endpoint_url/live/status?seconds=60`
`endpoint_url/live/status?meeting_id=123456789`
`endpoint_url/live/status?recording_id=abcdefg`

Requesting a number of seconds returns the last X seconds of updates.

Meeting id is the 9 or 10 digit meeting id.

Recording id is the uuid of the recording. This id must be url encoded in the request.

Attempts to use multiple identifiers, i.e. `endpoint_url/live/status?recording_id=abcdefg&meeting_id=123456789` will choose the most specific id (in this case the recording_id). If that's confusing, we could also just treat these requests as invalid.


functions/zoom-webhook.py
```diff
@@ -80,22 +82,48 @@ def handler(event, context):
...
+        origin = "on_demand"
+        correlation_id = payload["on_demand_request_id"]
+    else:
+        origin = "webhook_notification"
+        correlation_id = context.aws_request_id
+
     try:
         validate_payload(payload)
+        common.set_pipeline_status(
+            correlation_id, common.PipelineStatus.WEBHOOK_RECEIVED,
```
lbjay: A nitpicky thing i've picked up from gabe and the eslint rules is that once you break function args into multiple lines, you should do it uniformly, so that *every* arg goes on it's own line. We should try to fix these where we see them.

functions/zoom-webhook.py
```diff
@@ -80,22 +82,48 @@ def handler(event, context):
...
+            meeting_id=payload["object"]["id"],
+            recording_id=payload["object"]["uuid"],
+            origin=origin
+        )
     except BadWebhookData as e:
-        return resp_400("Bad data: {}".format(str(e)))
+        common.set_pipeline_status(
+            correlation_id, common.PipelineStatus.WEBHOOK_FAILED,
```
lbjay: each arg on a separate line

tests/test_downloader.py
```diff
@@ -89,6 +90,10 @@ def test_no_schedule_match(self):
         self.mocker.patch.object(
             downloader.Download, "_class_schedule", return_value={}
         )
+        mock_set_pipeline_status = self.mocker.Mock(return_value=None)
+        self.mocker.patch.object(
+            downloader, 'set_pipeline_status', mock_set_pipeline_status
```
lbjay: each arg on a separate line. not going to make the same comment for each instance. just fix where you see them.

lbjay (changes requested):
Several places where the formatting should be updated, but not finding any obvious problems with the code. However, the tests aren't passing for me. Is there maybe a `.env` setting that I'm missing? Example:
```
                    failure_msg = {"no_oc_series_found": dl_data}
                    common.set_pipeline_status(
>                       dl_data["correlation_id"],
                        common.PipelineStatus.IGNORED,
                        reason="No opencast series match"
                    )
E                   KeyError: 'correlation_id'
```

Something that would help me evaluate is a summary of how this solves the problem. Like, pretend I'm Kenny or a producer and explain how it's going to be easier to track my ingests.

PR #116, opened 2021-02-03
B/zip60 handling deleted file links

nataliemona:


functions/zoom-downloader.py
```diff
@@ -36,7 +37,7 @@ class PermanentDownloadError(Exception):
     pass
 
 
-class ZoomFileAccessError(Exception):
+class ZoomDownloadLinkError(Exception):
```
nataliemona, line 40: Changed exception name to be more general since we're giving up on telling the difference between types of download link failures (access vs deleted file, etc)

functions/zoom-downloader.py
```diff
@@ -129,11 +130,12 @@ class Download:
 
     def __init__(self, sqs, data):
         self.sqs = sqs
-        self.data = data
+        self.download_message = deepcopy(data)  # do not modify
```
nataliemona, line 133: We need a copy of the original message to send to the dlq. Dlq messages were failing because self.data was being modified to include a datetime object that was not JSON serializable ("created_local") but I don't want to make that into a string because we actually do want to use the datetime object in a few places and not be constantly converting back and forth.

functions/zoom-downloader.py
```diff
@@ -500,24 +503,8 @@ def zoom_filename(self):
 
             # If the file has been deleted or if the request is not authorized,
             # Zoom will return 200 and an HTML error page.
-            # This is bad, but we have no choice but to try to interpret the
-            # content that Zoom returns to us.
             if "Content-Type" in r.headers and "text/html" in r.headers["Content-Type"]:
-                # Most likely deleted file
```
nataliemona, line 506: All this code trying to figure out what is going on in the HTML page is too confusing and unreliable so I decided to just get rid of it and simplify things.

tests/test_downloader.py
```diff
@@ -402,6 +402,81 @@ def test_oc_series_found(self):
                 log_message = cm.output[-1].split(":")[-1]
                 assert log_message == expected_msg
 
+    def test_upload_to_s3(self):
```
nataliemona, line 405: Make sure we can actually successfully ignore a deleted file link and still have a successful ingest.

functions/zoom-downloader.py
```diff
@@ -320,7 +322,7 @@ def upload_message(self):
         if not hasattr(self, "self._upload_message"):
             s3_files = {}
             segment_durations = {}
-            for file in self.recording_files:
+            for file in self.downloaded_files:
```
nataliemona, line 325: Uploader only needs the metadata of the successfully downloaded files.

functions/zoom-downloader.py
lbjay: this reminds me of something i feel like i've lost sight of: we are creating the dead letter queues separate from the functions and then manually putting messages in them in the case of failures, but the lambda creation supports a DLQ as a [construct property](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda.Function.html#deadletterqueue) and will write events to it automatically (after retries expire). See https://aws.amazon.com/about-aws/whats-new/2016/12/aws-lambda-supports-dead-letter-queues/. So what is going on here? Was our implementation a workaround for something?

lbjay (approved):
This looks good. I just had one tangentially related question about the DLQs. Also, can you briefly summarize the strategy? It looks like we're going to just log an exception when we get a deleted file (or other "text/html" response) and move on with the ingest?

functions/zoom-downloader.py
nataliemona: It was for sending things to deadletter that we knew were not retryable... maybe it's not worth it though

PR #122, opened 2021-04-27
f/zip 63 status endpoint

nataliemona:
* Status endpoint & status table
* README explaining all ZIP endpoints (including status)
* Reorganization of common.py code into utils folder
* Change S3 bucket to have 1 week retention policy (from 2 weeks)
* Unit testing (75% coverage)
* Some refactoring. For example moving the schedule checking into common code so the future Slack app can share this function with the downloader lambda.

Post PR review updates:
* zip_id replaces "correlation_id" and "request_id"
* Add "webinar.ended" event to status tracking
* Remove API call to check for past meetings (missed cases were webinars?)
* Fix dynamo conditional update syntax


README.md
```diff
@@ -155,20 +157,85 @@ and enter the API endpoint under "Event Subscription."
 1. Make sure to subscribe to "All recordings have completed" events.
 1. Activate the app when desired. (For development it's recommended that you only leave the notifications active while you're actively testing.)
 
-### Setup on demand ingests
 
-The easiest way to find the on demand ingest endpoint is to run `invoke stack.status`.
+## Endpoints
```
lbjay, line 173: This is great. Nice addition.

cdk/api.py
```diff
@@ -78,61 +80,28 @@ def __init__(
...
-            method_responses=[
-                apigw.MethodResponse(
-                    status_code="200",
-                    response_models={
-                        "application/json": apigw.Model.EMPTY_MODEL
-                    },
-                )
-            ],
+        self.new_recording_resource = self.resource(
+            "new_recording", webhook_function, "POST"
```
lbjay: were we doing it wrong or did the cdk make this easier?

cdk/api.py
```diff
@@ -188,13 +164,40 @@ def endpoint_url(resource_name):
...
+        )
+
         core.CfnOutput(
             self,
             "RestApiId",
             export_name=f"{stack_name}-{names.REST_API}-id",
             value=self.api.rest_api_id,
         )
 
+    def resource(
```
lbjay: ahh I see. very good, but can this be called `create_resource`?

cdk/bucket.py
```diff
@@ -10,9 +10,9 @@ def __init__(self, scope: core.Construct, id: str):
         super().__init__(scope, id)
 
         two_week_lifecycle_rule = s3.LifecycleRule(
-            id="DeleteAfterTwoWeeks",
+            id="DeleteAfterOneWeek",
```
lbjay, line 13: should this be `one_week_lifecycle_rule` now?

functions/status-query.py
```diff
@@ -0,0 +1,64 @@
...
+    except InvalidStatusQuery as e:
+        return resp_400(e)
+
+    logger.info(records)
+
+    return {
+        "statusCode": 200,
+        "headers": {},
+        "body": json.dumps({"records": records}),
+    }
```
lbjay, line 64: what is the most number of records this could load and try to send back?

functions/utils/common.py
```diff
@@ -106,3 +127,53 @@ def zoom_api_request(
...
+    for event in schedule["events"]:
+        # match day
+        if zoom_day_code != event["day"]:
+            continue
+
+        # match time
+        scheduled_time = datetime.strptime(event["time"], "%H:%M")
+        timedelta = abs(
+            zoom_time
+            - zoom_time.replace(
```
lbjay: black does weird things sometimes

functions/utils/status.py
```diff
@@ -0,0 +1,299 @@
...
+        status_table = zip_status_table()
+        update_expression = (
+            "set update_date=:update_date, "
+            "update_time=:update_time, "
+            "expiration=:expiration, "
+            "pipeline_state=:pipeline_state"
+        )
+        expression_attribute_values = {
+            ":update_date": today,
+            ":update_time": int(seconds),
```
lbjay, line 97: Trying to understand why these are separate fields

functions/utils/status.py
```diff
@@ -0,0 +1,299 @@
...
+            UpdateExpression=update_expression,
+            ExpressionAttributeValues=expression_attribute_values,
+            ConditionExpression=condition_expression,
+        )
+    else:
+        status_table.update_item(
+            Key={"correlation_id": correlation_id},
+            UpdateExpression=update_expression,
+            ExpressionAttributeValues=expression_attribute_values,
+        )
```
lbjay: You don't have to change this, but just for comparison:

```
    update_args = dict(
        Key={"correlation_id": correlation_id},
        UpdateExpression=update_expression,
        ExpressionAttributeValues=expression_attribute_values,
    )
    if condition_expression:
        update_args["ConditionExpression"] = condition_expression

    status_table.update_item(**update_args)
```

functions/zoom-on-demand.py
```diff
@@ -61,31 +67,35 @@ def handler(event, context):
...
 
-    logger.info(f"Got recording uuid: '{uuid}'")
+    logger.info(f"Got recording uuid: '{recording_uuid}'")
 
     try:
         try:
             # zoom api can break if uuid is not double urlencoded
-            double_urlencoded_uuid = quote(quote(uuid, safe=""), safe="")
+            double_urlencoded_uuid = quote(
+                quote(recording_uuid, safe=""), safe=""
```
lbjay: do we care about keeping with the 1 arg per line thing?

functions/utils/status.py
lbjay: ok i get it now thanks to our slack convo. well, that 1MB data cap sure makes things interesting. do we know how many status update items will fit in 1MB? Hopefully a lot.

functions/utils/status.py
```diff
@@ -0,0 +1,299 @@
...
+                f":pipeline_state = {ZoomStatus.RECORDING_IN_PROGRESS.name} "
+                f"OR :pipeline_state = {ZoomStatus.RECORDING_PAUSED.name} "
+                f"OR :pipeline_state = {ZoomStatus.RECORDING_STOPPED.name} "
+            )
+        elif state in ZoomStatus:
+            # Enforce cannot transition back to a recording in progress state
+            # from recording processing
+            condition_expression = (
+                f":pipeline_state <> {ZoomStatus.RECORDING_PROCESSING.name}"
+            )
```
lbjay: I get this but also wonder what makes this a situation we have to protect against. Like, when would either of these cases happen?

functions/utils/status.py
```diff
@@ -0,0 +1,299 @@
...
+        if not meeting_id or not origin:
+            # When a recording enters the ZIP pipeline, for simplicity,
+            # only the first status tracking update includes additional metadata
+            # such as the meeting_id or origin. Subsequent status updates report
+            # status using a unique correlation id.
+            # Prevent adding records to dynamo status table for recordings
+            # that haven't been tracked since the beginning of the pipeline and
+            # therefore don't contain enough useful metadata. (This happens when
+            # you start status tracking for the first time or make modifications
+            # to the table that require it to be recreated.)
```
lbjay, line 134: my interpretation of this: each recording has only one db item. that db item can only be created when the recording 1st hits the pipeline via the webhook or on-demand endpoints. Subsequently it can be updated, but those update operations don't include the extra metadata, e.g., origin or meeting_id.

functions/zoom-on-demand.py
```diff
@@ -153,4 +168,13 @@ def handler(event, context):
...
 
+    set_pipeline_status(
+        request_id,
+        PipelineStatus.ON_DEMAND_RECEIVED,
+        meeting_id=webhook_data["payload"]["object"]["id"],
+        recording_id=recording_uuid,
+        recording_start_time=webhook_data["payload"]["object"]["start_time"],
+        topic=webhook_data["payload"]["object"]["topic"],
+        origin="on_demand",
+    )
```
lbjay: we are ok with not having a status entry if we returned a 500 error previous to this?

functions/zoom-webhook.py
```diff
@@ -41,6 +50,15 @@ def resp_400(msg):
     }
 
 
+STATUS_EVENT_TYPES = [
+    "recording.started",
+    "recording.resumed",
+    "recording.paused",
+    "recording.stopped",
+    "meeting.ended",
```
lbjay, line 57: These are new event types. Do you need to update the README section on webhook notification setup?

functions/zoom-webhook.py
```diff
@@ -101,6 +158,55 @@ def handler(event, context):
...
+    elif zoom_event == "recording.paused":
+        status = ZoomStatus.RECORDING_PAUSED
+    elif zoom_event == "recording.stopped":
+        double_urlencoded_uuid = quote(quote(uuid, safe=""), safe="")
+        r = zoom_api_request(
+            f"/past_meetings/{double_urlencoded_uuid}",
+            ignore_failure=True,
+        )
+        if r.status_code == 404:
+            status = ZoomStatus.RECORDING_STOPPED
```
lbjay: can you explain how/why a 404 translates to this status?

functions/zoom-webhook.py
```diff
@@ -101,6 +158,55 @@ def handler(event, context):
...
+            logger.info(f"Meeting {uuid} ended. Recording processing.")
+            status = ZoomStatus.RECORDING_PROCESSING
+    elif zoom_event == "meeting.ended":
+        if record_exists(correlation_id):
+            status = ZoomStatus.RECORDING_PROCESSING
+        else:
+            return resp_204(
+                f"Ignore meeting.ended for meeting id {mid} uuid {uuid} "
+                "not recorded"
+            )
```
lbjay: Are we sure this covers all the cases where we would get a meeting.ended event without already having an existing item? What about the "start status tracking for the first time" or table modification case?

functions/zoom-on-demand.py
```diff
@@ -134,6 +146,9 @@ def handler(event, context):
             "allow_multiple_ingests"
         ]
 
+    request_id = f"on-demand-{str(uuid4())}"
```
lbjay: this id value goes through a lot of naming transitions (request_id, on_demand_request_id, correlation_id) and correlation_id doesn't really capture what it represents any more. Should we just call it "zip_id"?

tasks.py
```diff
@@ -896,9 +902,12 @@ def __build_function(ctx, func, upload_to_s3=False):
             "pip install -U -r {} -t {}".format(req_file, build_path), hide=1
         )
 
-    for module in [func, "common", "gsheets"]:
-        module_path = join(dirname(__file__), "functions/{}.py".format(module))
-        module_dist_path = join(build_path, "{}.py".format(module))
+    mkdir(join(build_path, "utils"))
+    modules = ["utils/" + f.split(".")[0] for f in listdir("functions/utils")]
```
lbjay, line 906: Probably not a concern in this case, but for future reference `os.path.splitext` will do this "smarter" and handle a file like `foo.20210511.txt`

tests/test_uploader.py
```diff
@@ -73,6 +74,10 @@ def test_upload_count_ok(handler, mocker):
     mocker.patch.object(
         uploader, "get_current_upload_count", mocker.Mock(return_value=3)
     )
+    mock_set_pipeline_status = mocker.Mock(return_value=None)
+    mocker.patch.object(
+        uploader, "set_pipeline_status", mock_set_pipeline_status
+    )
```
lbjay: it bothers me that this mocking setup needs to be repeated in every test, so i'm playing around with some pytest stuff to see if I can work out how we could do it only once in a conftest fixture.

lbjay (commented):
I'm going to go ahead and submit this, but I still have an issue with the tests that I'm trying to work out.

tests/conftest.py
```diff
@@ -127,3 +142,39 @@ def _message_maker(frozen_time, zoom_event):
         return msg
 
     return _message_maker
+
+
+@pytest.fixture(autouse=True)
```
lbjay, line 148: There did not seem to be a way to create a single fixture. `mock.patch` only allows one target when used this way. `autouse` is nice because the mock gets activated for all functions. If you need to check call counts, etc, you can still pull the fixture into the test function scope by including it in the arg list (like I did in some of the webhook tests).

functions/utils/status.py
nataliemona: 1 MB is about 3,200 rows in the table which is a few days worth of recordings
So yes, a lot

functions/utils/status.py
nataliemona: In testing I observed "recording stopped" arriving after the "meeting ended" which then switched the recording state from "recording_processing" back to "recording_stopped"

functions/zoom-webhook.py
nataliemona: Recording stopped can translate to two statuses. Either "recording_stopped" if the meeting is still going on, or "recording_processing" if the meeting has ended. 

Getting 404 on the "past_meetings" endpoint means that the meeting has not ended yet and so we should use the "recording_stopped" status rather than "recording_processing"

functions/zoom-webhook.py
nataliemona: The issue is that meeting ended doesn't say anything about whether there was a recording or not. So we really don't want to add it into tracking unless we have a way of knowing that it was a recorded meeting. There are significantly more non-recorded meetings than there are recorded meetings.

functions/status-query.py
nataliemona: For meeting_id query, that would be the number of recordings in a week for a single meeting which is probably somewhat limited. Probably most commonly <4. Maybe 20 on the extremely high outlier side.
For recent updates, if currently it only pulls updates from the last hour than that's probably not more than 100-200 (estimate).

functions/utils/common.py
nataliemona: rewriting this a bit to make it easier to read

functions/zoom-on-demand.py
nataliemona: I guess my thinking on this was that the requestor would know that their request failed because they would see an error in Opencast. Otherwise you might see a lot of failed requests in the status tools like Slack, etc. But maybe we want that data?

PR #126, opened 2021-06-02
f/zip 43 slack app

nataliemona:
Slack integration.
- Additional endpoint for Slack app.
- Instructions for setting up Slack app.
- Test coverage 77%.


cdk/api.py
```diff
@@ -104,6 +105,12 @@ def __init__(
             "status", status_query_function, "GET"
         )
 
+        self.slack_resource = self.create_resource(
+            "slack",
+            slack_function,
+            "POST",
+        )
+
```
lbjay: can you maybe add the magic commas to the `create_resoruce` calls above this one so that it's easy to tell at a glance they're all doing the same thing?

cdk/app.py
```diff
@@ -62,6 +62,7 @@
     "default_publisher": default_publisher,
     "override_publisher": getenv("OVERRIDE_PUBLISHER", required=False),
     "override_contributor": getenv("OVERRIDE_CONTRIBUTOR", required=False),
+    "oc_cluster_name": getenv("OC_CLUSTER_NAME"),
```
lbjay, line 65: This is a test comment using the vscode github pull request extension

cdk/schedule.py
```diff
@@ -1,23 +0,0 @@
-from aws_cdk import core, aws_dynamodb as dynamodb
```
lbjay, line 1: Trying to figure out why this was deleted

function_requirements/status-query.txt
```diff
@@ -6,27 +6,27 @@
 #
```
lbjay, line 6: Just a side effect that these reqs all got updates?

functions/zoom-downloader.py
```diff
@@ -462,7 +462,12 @@ def __init__(self, file_data, segment_num):
         self.recording_type = self.__standardized_recording_type(
             file_data["recording_type"]
         )
-        self.s3 = boto3.client("s3")
+
+    @property
```
lbjay: i'm sure this is a good change but i'm curious about what prompted it. it also made me wonder what would happen if the s3 client was a class attribute (vs instance attribute).

functions/zoom-on-demand.py
```diff
@@ -177,5 +177,6 @@ def handler(event, context):
         recording_start_time=webhook_data["payload"]["object"]["start_time"],
         topic=webhook_data["payload"]["object"]["topic"],
         origin="on_demand",
+        oc_series_id=body["oc_series_id"] if "oc_series_id" in body else None,
```
lbjay: why not just `body.get("oc_series_id")`?

lbjay (approved):
This looks good to me. Can we be sure to squash into one commit before merging?

function_requirements/status-query.txt
nataliemona: Was updating urllib3 to resolve dependabot complaint

functions/zoom-downloader.py
nataliemona: Purpose was for unit testing - allow creating a ZoomFile without creating an s3 client. But could probably be done in a better way...

cdk/schedule.py
nataliemona: Should have been deleted in previous change (status endpoint PR)

functions/zoom-downloader.py
nataliemona: I ended up putting the boto3 clients/resources at the top of the file

PR #130, opened 2021-06-18
bug fix - Sneaky file param generator bug.

nataliemona:
Really sneaky bug. Thank you @rute-santos for the find.

Because `self._used_views.append(view)` was indented and inside the s3_filenames loop, and also because `self._used_views` was a list and not a set. The `self._used_views` contained as many items as the number of files, not the number of views.

For example, ingesting a 3 segment recording with the views "active_speaker" and "shared_screen_with_speaker" results in the `self._used_views` list `["active_speaker", "active_speaker", "active_speaker"]`.

When the check `if len(self._used_views) >= len(self.FLAVORS):` is done when attempting to add "shared_screen_with_speaker" as the secondary view, the length of `self._used_views` is 3, which is equal to the number of flavors, and the secondary view is skipped.

### DIFF
--- a/functions/zoom-uploader.py
+++ b/functions/zoom-uploader.py
@@ -445,7 +445,7 @@ class FileParamGenerator(object):
 
     def __init__(self, s3_filenames):
         self.s3_filenames = s3_filenames
-        self._used_views = []
+        self._used_views = set()
         self._params = []
         # whatever the max length of any view's list of files is the number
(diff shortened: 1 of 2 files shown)


lbjay (approved):
Yikes. Also, how did I get away with having both a `self.flavors` and `self.FLAVORS`. smh.

PR #131, opened 2021-06-22
T/zip 64 download tokens

nataliemona:
Switch from using zak token to using download token.


functions/zoom-uploader.py
```diff
@@ -373,8 +373,9 @@ def get_series_catalog(self):
         try:
             resp = oc_api_request("GET", endpoint)
         except requests.exceptions.HTTPError as e:
-            if e.response.status_code == "404":
+            if e.response.status_code == 404:
```
nataliemona, line 376: Status code is an integer... not a string... whoops

functions/slack/handler.py
```diff
@@ -263,7 +263,7 @@ def slack_api_request(endpoint):
         f"https://slack.com/api/{endpoint}",
         headers={"Authorization": f"Bearer {SLACK_API_TOKEN}"},
     )
-    r.raise_for_status
+    r.raise_for_status()
```
lbjay, line 266: there ought to be a pylint warning for this exact bug

tests/conftest.py
```diff
@@ -114,6 +115,8 @@ def _message_maker(frozen_time, payload):
             zip_id = payload["payload"]["zip_id"]
         else:
             zip_id = f"auto-ingest-{payload_obj['uuid']}"
+
+        on_demand_ingest = True if zoom_event == "on.demand.ingest" else False
```
lbjay: Hmmm. Why not 
```
on_demand_ingest = zoom_event == "on.demand.ingest"
```
?

lbjay (changes requested):
Just one small change

PR #138, opened 2021-10-08
add `BUFFER_MINUTES` to downloader function environment

lbjay:
Our `.env` file has a `BUFFER_MINUTES` setting, and the downloader function will read a `BUFFER_MINUTES` environment variable, but the CDK code doesn't actually create a `BUFFER_MINUTES` environment variable for the function. The function relies on a default setting of 30 minutes, which was fine except now we want to be able to adjust the `BUFFER_MINUTES` setting.


rute-santos (approved):
Looks good to me!

lbjay:
added a test for the buffer minutes matching stuff

PR #139, opened 2021-10-15
ZIP-74: Ingest chat files to Opencast

rute-santos:
- Chat files ingested as attachment with flavot chat/chunked+source.
- Send workflow configuration with start/end times of all segments ingested so that Opencast can adjust the times (chat message times are relative to the event start time, not recording times, and recoding can be stopped/started during an event).

### DIFF
--- a/functions/zoom-uploader.py
+++ b/functions/zoom-uploader.py
@@ -178,6 +178,7 @@ def process_upload(upload_data):
 class Upload:
     def __init__(self, data):
         self.data = data
+        self.recording_times = ""
 
     @property
     def created(self):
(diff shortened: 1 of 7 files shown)


functions/zoom-uploader.py
```diff
@@ -292,19 +293,50 @@ def s3_filenames(self):
...
         :return: [description]
         :rtype: [type]
         """
 
         if not hasattr(self, "_s3_filenames"):
             self._s3_filenames = {}
+            segment_0_discarded = False
+            chat_has_segment_0 = False
+            # Key = seg number, value = "recordingStart_recordingEnd"
+            rec_times = {}
```
lbjay, line 312: is there a reason this is a dictionary as opposed to just a list where the `segment_num` corresponds to the array index?

lbjay (approved):
looks good!

PR #140, opened 2022-03-08
Don't assume `ingest_request_time` is present

lbjay:


functions/slack/blocks.py
```diff
@@ -209,8 +209,10 @@ def ingest_details(rec, schedule):
             )
             on_demand = ingest["origin"] == "on_demand"
 
-            request_time = formatted_local_time(ingest["ingest_request_time"])
-            if on_demand:
+            if on_demand and "ingest_request_time" in ingest:
```
rute-santos: I don't think this is only for on_demand. See: https://github.com/harvard-dce/zoom-recording-ingester/blob/b62606015930c9cc26144577b57859a8960a12f9/functions/utils/status.py#L105

lbjay: Yeah, it's weird. That value isn't accessed anywhere else so I figured this was safe, but I think I'm being lazy about it and should really try to track down how/why that value is missing sometimes. Don't approve this. I'll go back and look a little deeper.

rute-santos: It's missing when we don't get the webhook notification (at least in the case I looked at).

rute-santos (approved):

PR #143, opened 2022-08-24
ZIP-81 OC11: use the external api event endpoint instead of workflow endpoint when checking existence of mp

rute-santos:


lbjay (approved):
Looks great

PR #151, opened 2023-05-30
Jluker remove op counts func

lbjay:


rute-santos (approved):
This looks good. Does the README need any updates?

PR #152, opened 2023-06-12
Migrate from CDK v1 to v2

lbjay:
- Initial version of aws-cdk-lib is 2.81.0
- Reworking the imports of core cdk classes and constructs
- Minor changes to how lambda versions/aliases are defined
- Minor rearranging of arguments to cloudwatch Alarm and Metric constructs
- We now include a `package.json` for installing the aws-cdk CLI with the matching version
- tasks that invoked `cdk` now run `npx cdk` to ensure we're running the locally installed cli


rute-santos (approved):
Looks good to me.

PR #153, opened 2023-06-12
Implements new Zoom webhook endpoint validation mechanism

lbjay:
- New env var: `WEBHOOK_VALIDATION_SECRET_TOKEN`
- zoom-webhook function now accepts "endpoint.url_validation" events and will process and return the apprpriate hashed token value


rute-santos (approved):
Looks good!

PR #154, opened 2023-07-24
ZIP-89: remove zoom api call to get `host_name` value

lbjay:
Trying to minimize the zoom api "scopes" required. This value is passed into the SQS message to the uploader but never used


rute-santos (approved):
Looks ok to me!

PR #158, opened 2023-09-20
ZIP-91 Saving webhook events recording.paused/resumed/started/stopped to a db

rute-santos:
This introduces a separate dynamodb table to hold a list of recording.paused/resumed/started/stopped events.
I could have added this to the status table but I wanted a longer expiration (4 months, which covers at least a term). Please let me know your thoughts about this.
The timestamps recorded are sent to Opencast at ingest time as a workflow configuration (before this PR, it would only send start/stop times gotten from the recordings). This is so that Opencast can properly synchronize the chat to the video.


lbjay:
The README has a section that lists which zoom events to subscribe to, so you'll need to add the new ones there.

cdk/recording_events_table.py
```diff
@@ -0,0 +1,33 @@
+from aws_cdk import Stack, RemovalPolicy, aws_dynamodb as dynamodb
+from constructs import Construct
+from . import names
+
+
+class RecordingEvents(Construct):
```
lbjay: This should be `ZipRecordingEvents` for consistency with other construct names.

cdk/recording_events_table.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, scope: Construct, id: str):
+        """
+        This table records the times of some webhook events.
+        """
+        super().__init__(scope, id)
+        stack_name = Stack.of(self).stack_name
+
+        self.table = dynamodb.Table(
```
lbjay, line 14: I see there is now a `dynamodb.TableV2` class that is recommended. Looks like it's very new, so IMO it's fine to stick with the old `Table` here. Just mentioning it.

cdk/recording_events_table.py
```diff
@@ -0,0 +1,33 @@
...
+            self,
+            "table",
+            table_name=f"{stack_name}-{names.RECORDING_EVENTS_TABLE}",
+            partition_key=dynamodb.Attribute(
+                name="zoom_uuid",
+                type=dynamodb.AttributeType.STRING,
+            ),
+            billing_mode=dynamodb.BillingMode.PAY_PER_REQUEST,
```
lbjay, line 22: Did you look into this or just going on intuition?

cdk/recording_events_table.py
```diff
@@ -0,0 +1,33 @@
...
+            removal_policy=RemovalPolicy.DESTROY,
+            time_to_live_attribute="expiration",
+        )
+
+        self.table.add_global_secondary_index(
```
lbjay: Can you explain the need for this secondary index? I'm re-reading to refresh my memory of dynamodb so maybe i'm missing something.

functions/zoom-uploader.py
```diff
@@ -298,10 +299,10 @@ def s3_filenames(self):
                 # the video.
                 for seg in segment_files:
```
lbjay, line 300: Was there alread some unused code in here that was meant for ZIP-74?

functions/utils/recording_events.py
```diff
@@ -0,0 +1,245 @@
...
+from os import getenv as env
+from dotenv import load_dotenv
+from os.path import join, dirname
+from .common import TIMESTAMP_FORMAT
+from pytz import utc
+from itertools import chain
+
+logger = logging.getLogger()
+
+load_dotenv(join(dirname(__file__), "../.env"))
```
lbjay, line 14: it's weird that all these `utils` modules load the dotenv file themselves. Not sure how/why this got structured this way.

functions/utils/recording_events.py
```diff
@@ -0,0 +1,245 @@
...
+
+RECORDING_EVENTS_TABLE = env("RECORDING_EVENTS_TABLE", None)
+
+
+RECORDING_EVENTS = [
```
lbjay: If you put this a few lines later it could be 
`RECORDING_EVENTS = START_EVENTS + END_EVENTS`

functions/utils/recording_events.py
```diff
@@ -0,0 +1,245 @@
...
+    "recording.resumed",
+    "recording.stopped",
+]
+
+START_EVENTS = ["recording.started", "recording.resumed"]
+END_EVENTS = ["recording.paused", "recording.stopped"]
+
+# This was copied from status.py:
+# If RECORDING_EVENTS_TABLE is not set we assume the events table
+# is not being used. (For example tasks.py imports utils but doesn't
```
lbjay, line 24: More unfortunate weirdness with how the `utils` modules get used

cdk/recording_events_table.py
lbjay: oh i guess if we used `PROVISIONED` it could get used up and that would be bad?

functions/utils/recording_events.py
```diff
@@ -0,0 +1,245 @@
...
+        "timestamp": "TIMESTAMP_IN_SECS",
+    }
+    """
+    # If not one of desired events or no timestamp, ignore
+    if zoom_event not in RECORDING_EVENTS or not zoom_event_timestamp:
+        return
+
+    # If this is an event for recording started/paused/resumed/stopped,
+    # record the event time in the table in a list.
+    formatted_local_time = datetime.strftime(
```
lbjay, line 64: what's "local" about this?

functions/zoom-uploader.py
lbjay: oh nevermind. ZIP-74 is older. This is ZIP-91.

functions/zoom-uploader.py
```diff
@@ -331,7 +332,15 @@ def s3_filenames(self):
                     ][1:]
 
             # ZIP-74 "start1_end1,start2_end2,start3_end3"
```
lbjay, line 334: Is this still the format?

functions/utils/recording_events.py
```diff
@@ -0,0 +1,245 @@
...
+            ")"
+        )
+        # Expiration 120 days, which should cover at least a semester
+        expression_attribute_values = {
+            ":expiration": int((utcnow + timedelta(days=120)).timestamp()),
+            ":new_recording_event": new_recording_event_list,
+            ":empty_list": [],
+        }
+
+        if meeting_id:
```
lbjay: When would there not be a `meeting_id`? Also why is it included in the item? I don't see it being referenced in queries.

cdk/recording_events_table.py
rute-santos: I just copied it from the other tables...

cdk/recording_events_table.py
rute-santos: I didn't look up; I copied from the status table because it's similar.

cdk/recording_events_table.py
rute-santos: No need. I can remove it.

functions/zoom-uploader.py
rute-santos: Yes, same format as before.

functions/utils/recording_events.py
rute-santos: It isn't. I can remove it.

functions/utils/recording_events.py
rute-santos: Local time instead of utc, just for clarity.

cdk/recording_events_table.py
rute-santos: I think on-demand is best because it's unpredictable traffic and we want to pay only for what we use (https://docs.aws.amazon.com/wellarchitected/latest/serverless-applications-lens/capacity.html)

rute-santos:
Hi @lbjay , Thanks for the review! I think I addressed all your comments, but please let me know if there's anything else.

functions/utils/recording_events.py
lbjay: oh ok. you're purposely not converting to UTC. That's fine for the purposes of logging something, but the comments just above are misleading.

functions/utils/recording_events.py
```diff
@@ -0,0 +1,233 @@
...
+
+# This was copied from status.py:
+# If RECORDING_EVENTS_TABLE is not set we assume the events table
+# is not being used. (For example tasks.py imports utils but doesn't
+# use the recording times table)
+recording_events_table = None
+if RECORDING_EVENTS_TABLE:
+    dynamodb = boto3.resource("dynamodb")
+    recording_events_table = dynamodb.Table(RECORDING_EVENTS_TABLE)
+
```
lbjay: It took me a bit but I think I finally got it. The `RECORDING_EVENTS_TABLE` variable will only be set in the context of a running lambda. Maybe update the comment to something more clear, like:

```
This value only ever comes from the lambda function's environment variable configuration. In a local environment, i.e., when running `tasks.py`, it will not be set.
```

rute-santos: To be honest, I just copied it from Mona's code without really understanding why. The previous comment was not very helpful :) I will add the comment you suggested.

functions/utils/recording_events.py
rute-santos: I've changed the comment to make it clear.

lbjay (approved):

PR #163, opened 2024-07-15
Add retries to the individual streaming of files from zoom to s3

lbjay:
- added tenacity (retry library) to function requirements
- failed files will be retried up to 5 times with a 2m wait in between


rute-santos (approved):
Looks good.

functions/zoom-downloader.py
```diff
@@ -358,8 +367,22 @@ def upload_to_s3(self):
...
         for file in self.recording_files:
             try:
-                file.stream_file_to_s3()
+                retry_attempts = Retrying(
+                    stop=stop_after_attempt(STREAM_FROM_ZOOM_TO_S3_RETRIES),
+                    wait=wait_fixed(STREAM_FROM_ZOOM_TO_S3_RETRY_WAIT),
+                    retry=retry_if_exception_type(RetryableDownloadError),
+                )
+                for attempt in retry_attempts:
+                    with attempt:
```
rute-santos, line 378: It would be helpful to log something here for debugging purposes...

karendolan:
Hi @lbjay I saw a few recommendations for preventing chunk transfer latency timeouts by adding the "tcp_keepalive=True" to the boto3 client. The default is false. I added notes on the ZIP-93 ticket. In summary:

Consider changing
`s3 = boto3.client("s3")` 
to
`s3 = boto3.client("s3", config=Config(tcp_keepalive=True))`
In https://github.com/harvard-dce/zoom-recording-ingester/blob/main/functions/zoom-downloader.py#L51

That param is available in the version of boto3 used in this repo (1.26.142) https://github.com/boto/boto3/blob/1.26.142/docs/source/guide/configuration.rst#tcp_keepalive

Here is an example of ensuing the param is supported https://github.com/brightbock/cloudwatch-logs-tag/blob/main/src/lambda.py#L192-L201

PR #168, opened 2024-12-09
use a new Session for each request

lbjay:
this attempts to work around an issue we see with requests to Opencast possibly failing when re-using urllib https connections

### DIFF
--- a/functions/zoom-uploader.py
+++ b/functions/zoom-uploader.py
@@ -41,6 +41,7 @@ aws_lambda = boto3.client("lambda")
 sqs = boto3.resource("sqs")
 
 UPLOAD_OP_TYPES = ["track", "uri-track"]
+SESSION = None
 
 
 class OpencastConnectionError(Exception):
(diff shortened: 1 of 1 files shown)


functions/zoom-uploader.py
```diff
@@ -40,19 +40,6 @@
...
 
-session = requests.Session()
-session.auth = HTTPDigestAuth(OPENCAST_API_USER, OPENCAST_API_PASSWORD)
-session.headers.update(
-    {
-        "X-REQUESTED-AUTH": "Digest",
-        # TODO: it's possible this header is not necessary for the endpoints
-        # being used here. It seems like for Opencast endpoints where the
-        # header *is* necessary the correct value is actually
-        # "X-Opencast-Matterhorn-Authorization"
```
lbjay, line 51: i just removed the comment, but maybe we should deal with this confusion too.

functions/zoom-uploader.py
```diff
@@ -67,6 +54,16 @@ class InvalidOpencastSeriesId(Exception):
...
+
+    session = requests.Session()
+    session.auth = HTTPDigestAuth(OPENCAST_API_USER, OPENCAST_API_PASSWORD)
+    session.headers.update(
+        {
+            "X-REQUESTED-AUTH": "Digest",
+            "X-Opencast-Matterhorn-Authentication": "true",
+        }
+    )
+
```
rute-santos: This will create a new session for each request. It would be preferable to have one session per run, right? How about creating the session in process_upload or Upload init?

lbjay: ok i redid it so that a Session gets created (and closed) once per invocation

rute-santos (approved):

PR #171, opened 2025-05-16
ZIP-99 Update Python and dependencies

rute-santos:
Python was updated to 3.12.
All dependencies were updated (hopefully).
This cause an increase in the size of the downloader and on-demand functions, which was fixed by always uploading the code to S3 and updating the function using the S3 content.


requirements/tox.txt
```diff
@@ -1,166 +1,161 @@
...
     #   google-auth
-pyjwt==2.7.0
-    # via -r requirements/../function_requirements/common-requirements.txt
-pyparsing==3.0.9
+pyjwt==2.10.1
+    # via -r /Users/ras300/Documents/opencast-16.x/zoom-ingester-oc-rute-16x/function_requirements/common-requirements.txt
+pyparsing==3.2.3
     # via
-    #   -r requirements/../function_requirements/common-requirements.txt
+    #   -r /Users/ras300/Documents/opencast-16.x/zoom-ingester-oc-rute-16x/function_requirements/common-requirements.txt
```
lbjay, line 104: These are just annotation comments but it looks strange. I wonder why it's doing this.

functions/utils/status.py
```diff
@@ -154,9 +154,9 @@ def set_pipeline_status(
                 "OR pipeline_state = :recording_stopped "
             )
         elif state in ZoomStatus:
-            expression_attribute_values[
-                ":recording_processing"
-            ] = ZoomStatus.RECORDING_PROCESSING.name
+            expression_attribute_values[":recording_processing"] = (
```
lbjay, line 157: This looks odd. Was it a formatting change made by black?

lbjay (approved):
I think this is good. Just had one question about the requirements files annotations.

PR #174, opened 2025-06-27
ZIP-101 Fix exception in uploader when recording is too short

rute-santos:
Abort ingesting of short recording in downloader.
If we want to allow override when ingesting on-demand, we need to provide an explicit way to choose that.

### DIFF
--- a/functions/utils/recording_events.py
+++ b/functions/utils/recording_events.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, UTC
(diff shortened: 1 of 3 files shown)


functions/utils/recording_events.py
```diff
@@ -235,4 +235,6 @@ def from_utc_to_timestamp(date_str):
 
 
 def from_timestamp_to_utc(ts):
```
rute-santos, line 237: This is a commit from ZIP-100

lbjay (approved):

